PR #42, opened 2022-06-27
✨ add cloudbuild config for releases

charlie-haley:
This PR adds cloudbuild.yaml config for building releases. It's based on configuration from existing cluster-api providers


alexander-demichev (approved):
/ok-to-test
/approve
/lgtm

PR #72, opened 2023-02-27
🌱 Enable additional linters

Fedosin:

To improve the code quality we enable additional golangci linters. Some desired linter are commented in `.golangci.yaml` file, but they will be enabled in the future.


alexander-demicev (approved):
/approve
/lgtm

PR #81, opened 2023-03-13
:book: Replace apiversion from management to operator

ljtill:

Update documentation for apiVersion values to `operator.cluster.x-k8s.io/v1alpha1`


alexander-demicev (approved):
Thanks a lot!
/lgtm
/approve

PR #89, opened 2023-03-21
🌱 Stop using github.com/pkg/errors

Fedosin:

This project is deprecated and not maintained anymore: https://github.com/pkg/errors

We can use standard golang primitives for that.


alexander-demicev:
@neolit123 thanks for the information. We will deprecate pkg/errors and see how well it works for us, good to know that people are still using it.

alexander-demicev (approved):
/approve
/lgtm

PR #90, opened 2023-03-21
🌱 Bump golangci-lint to v1.51.2

Fedosin:
New version supports golang 1.20


damdo (commented):
/lgtm

PR #102, opened 2023-04-24
🐛 Use correct flags in the component customizer

Fedosin:

In all CAPI componets the manager flags `--metrics-addr` and `enable-leader-election` were deprecated in favor of --`metrics-bind-addr` and `--leader-elect` to be more aligned with core Kubernetes Components.

https://github.com/kubernetes-sigs/cluster-api/pull/3934

All cluster api providers have made related changes. It leads to the fact that when we try to use old flags there, operator returns an error.

Examples of flag names:
[CAPI](https://github.com/kubernetes-sigs/cluster-api/blob/main/main.go#L138-L142), [Azure](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/main.go#L124-L136), [AWS](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/main.go#L422-L434), [GCP](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/main.go#L264-L276)


alexander-demicev (approved):
/lgtm
/approve

PR #105, opened 2023-05-05
🐛 fix: ensure that all required env variables are set before running tests

Fedosin:

Currently we don't set defaults for 2 mandatory env variables: CERT_MANAGER_VERSION and OPERATOR_IMAGE. If user doesn't define them manually before running tests, the execution will fail with some vague error messages.

This commit ensures that the env variables always have default values.


damdo (commented):
/lgtm

alexander-demicev (approved):
/approve
/lgtm

PR #107, opened 2023-05-08
✨ support provider downgrades

Fedosin:

Previously, the operator allowed only installation of new provider versions. Downgrading to a previous version would result in operator failure.

This PR introduces support for seamless provider downgrades in addition to upgrades.

### DIFF
--- a/internal/envtest/environment.go
+++ b/internal/envtest/environment.go
@@ -293,6 +293,33 @@ func (e *Environment) CreateNamespace(ctx context.Context, generateName string) 
 	return ns, nil
 }
 
+func (e *Environment) EnsureNamespaceExists(ctx context.Context, namespace string) error {
+	// Check if the namespace exists
+	ns := &corev1.Namespace{}
+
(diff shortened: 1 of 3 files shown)


internal/envtest/environment.go
```diff
@@ -293,6 +293,34 @@ func (e *Environment) CreateNamespace(ctx context.Context, generateName string)
...
+
+	// Create the namespace if it doesn't exist
+	newNamespace := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: namespace,
+		},
+	}
+
+	err = e.Client.Create(ctx, newNamespace)
+	if err != nil {
```
alexander-demicev, line 317:
```suggestion
	if err: = e.Client.Create(ctx, newNamespace); err != nil {
```

alexander-demicev, line 317:
```suggestion
	if err: = e.Client.Create(ctx, newNamespace); err != nil {
```

alexander-demicev (approved):
/approve

PR #108, opened 2023-05-08
🐛 Avoid failures on unchanged provider version

Fedosin:
In the current implementation, when a provider version remains the same during subsequent reconciliations, the operator encounters an error with the message "failed getting clusterctl Provider version," leading to a degraded status.

This PR addresses this issue by adding a check to determine whether the provider version has changed or not.

Fixes #


alexander-demicev (approved):
/approve

alexander-demicev (approved):
/lgtm

PR #111, opened 2023-05-15
✨ support gitlab repos

Fedosin:

clusterctl [supports](https://cluster-api.sigs.k8s.io/clusterctl/provider-contract.html#creating-a-provider-repository-on-gitlab) provider installation from gitlab repositories, so we should implement it as well.


alexander-demicev (approved):
/approve
/lgtm

PR #115, opened 2023-05-24
🐛 Don't requeue provider if its spec is invalid

Fedosin:

Currently if provider fails spec validation during preflight checks, we requeue it again with a 30 seconds interval. It doesn't make sense, because the spec still will be invalid and validation will fail again and again.

This PR identifies all places where we don't need to requeue providers and updates them accordingly.

### DIFF
--- a/internal/controller/preflight_checks.go
+++ b/internal/controller/preflight_checks.go
@@ -61,7 +61,7 @@ func preflightChecks(ctx context.Context, c client.Client, provider genericprovi
 			emptyVersionMessage,
 		))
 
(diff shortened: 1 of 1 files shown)


furkatgofurov7:
Presubmit job failures are not related to this PR but the issue with k8s test-infra itself (same in many other PRs' in different repos currently) and tracked in https://github.com/kubernetes/test-infra/issues/29622

Fedosin:
@furkatgofurov7 thanks! good to know

furkatgofurov7 (commented):
This definitely makes sense to me!

LGTM

furkatgofurov7 (commented):
/lgtm

alexander-demicev (approved):
Makes sense to me too, we discussed it with @Fedosin earlier this week.
/approve

PR #116, opened 2023-05-28
🌱 Bump kube-rbac-proxy

Fedosin:

Currently we use an outdated version of kube-rbac-proxy and don't specify resources requests and limits for it.

This PR bumps kube-rbac-proxy to the latest stable version and updates manifests based on the related recommendations.


alexander-demicev (approved):
/approve
/lgtm

PR #118, opened 2023-05-29
🌱 Configure importas linter and fix import issues 

furkatgofurov7:

Adds aliases for most common (k8s.io/*) and operator specific packages to make broader use of `importas` linter from the .golangci.yaml. Also, due to mandating of the imports, fixes some imports to pass the CI


Fedosin (approved):
/lgtm
/approve

PR #121, opened 2023-05-30
🌱 add more e2e tests

Fedosin:

This PR adds several e2e tests to verify deletion, upgrading and downgrading providers.

### DIFF
--- a/test/e2e/minimal_configuration.go
+++ b/test/e2e/minimal_configuration.go
@@ -22,14 +22,15 @@ package e2e
 import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	operatorv1 "sigs.k8s.io/cluster-api-operator/api/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var _ = Describe("Create providers with minimal specified configuration", func() {
-	It("should succefully create a CoreProvider", func() {
+var _ = Describe("Create, upgrade, downgrade and delete providers with minimal specified configuration", func() {
+	It("should successfully create a CoreProvider", func() {
 		k8sclient := bootstrapClusterProxy.GetClient()
 		coreProvider := &operatorv1.CoreProvider{
 			ObjectMeta: metav1.ObjectMeta{
@@ -85,7 +86,7 @@ var _ = Describe("Create providers with minimal specified configuration", func()
 		}, timeout).Should(Equal(true))
 	})
 
-	It("should succefully create a BootstrapProvider", func() {
+	It("should successfully create and delete a BootstrapProvider", func() {
 		k8sclient := bootstrapClusterProxy.GetClient()
 		bootstrapProvider := &operatorv1.BootstrapProvider{
 			ObjectMeta: metav1.ObjectMeta{
@@ -139,9 +140,22 @@ var _ = Describe("Create providers with minimal specified configuration", func()
 			}
 			return false
 		}, timeout).Should(Equal(true))
+
+		Expect(k8sclient.Delete(ctx, bootstrapProvider)).To(Succeed())
+
+		By("Waiting for the bootstrap provider deployment to be deleted")
+		Eventually(func() bool {
+			deployment := &appsv1.Deployment{}
+			key := client.ObjectKey{Namespace: operatorNamespace, Name: bootstrapProviderDeploymentName}
+			isBootstrapProviderReady, err := waitForObjectToBeDeleted(k8sclient, ctx, key, deployment)
+			if err != nil {
+				return false
+			}
+			return isBootstrapProviderReady
+		}, timeout).Should(Equal(true))
 	})
 
-	It("should succefully create a ControlPlaneProvider", func() {
+	It("should successfully create and delete a ControlPlaneProvider", func() {
 		k8sclient := bootstrapClusterProxy.GetClient()
 		cpProvider := &operatorv1.ControlPlaneProvider{
 			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


test/e2e/minimal_configuration.go
```diff
@@ -195,9 +209,22 @@ var _ = Describe("Create providers with minimal specified configuration", func()
...
+
+		By("Waiting for the control plane provider deployment to be deleted")
+		Eventually(func() bool {
+			deployment := &appsv1.Deployment{}
+			key := client.ObjectKey{Namespace: operatorNamespace, Name: cpProviderDeploymentName}
+			isReady, err := waitForObjectToBeDeleted(k8sclient, ctx, key, deployment)
+			if err != nil {
+				return false
+			}
+			return isReady
```
furkatgofurov7, line 223: how about:
```suggestion
			isCPProviderDeleted, err := waitForObjectToBeDeleted(k8sclient, ctx, key, deployment)
			if err != nil {
				return false
			}
			return isCPProviderDeleted
```
instead?

test/e2e/minimal_configuration.go
```diff
@@ -139,9 +140,22 @@ var _ = Describe("Create providers with minimal specified configuration", func()
 			}
 			return false
 		}, timeout).Should(Equal(true))
+
+		Expect(k8sclient.Delete(ctx, bootstrapProvider)).To(Succeed())
```
furkatgofurov7, line 144: Can we also update the line 89 in that case?

```
		It("should successfully create and delete a BootstrapProvider", func() {
```

test/e2e/minimal_configuration.go
```diff
@@ -22,14 +22,15 @@ package e2e
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	operatorv1 "sigs.k8s.io/cluster-api-operator/api/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var _ = Describe("Create providers with minimal specified configuration", func() {
-	It("should succefully create a CoreProvider", func() {
+var _ = Describe("Create, upgrade, downgrade and delete providers with minimal specified configuration", func() {
+	It("should successfully create a CoreProvider", func() {
```
furkatgofurov7, line 33: Do we want to add a delete test for CoreProvider as well?

test/e2e/minimal_configuration.go
```diff
@@ -251,5 +278,146 @@ var _ = Describe("Create providers with minimal specified configuration", func()
...
+
+		By("Waiting for the infrastructure provider deployment to be deleted")
+		Eventually(func() bool {
+			deployment := &appsv1.Deployment{}
+			key := client.ObjectKey{Namespace: operatorNamespace, Name: infraProviderDeploymentName}
+			isReady, err := waitForObjectToBeDeleted(k8sclient, ctx, key, deployment)
+			if err != nil {
+				return false
+			}
+			return isReady
```
furkatgofurov7, line 292:
```suggestion
			isInfraProviderDeleted, err := waitForObjectToBeDeleted(k8sclient, ctx, key, deployment)
			if err != nil {
				return false
			}
			return isInfraProviderDeleted
```

test/e2e/minimal_configuration.go
```diff
@@ -139,9 +140,22 @@ var _ = Describe("Create providers with minimal specified configuration", func()
...
+
+		By("Waiting for the bootstrap provider deployment to be deleted")
+		Eventually(func() bool {
+			deployment := &appsv1.Deployment{}
+			key := client.ObjectKey{Namespace: operatorNamespace, Name: bootstrapProviderDeploymentName}
+			isReady, err := waitForObjectToBeDeleted(k8sclient, ctx, key, deployment)
+			if err != nil {
+				return false
+			}
+			return isReady
```
furkatgofurov7, line 154:
```suggestion
			isBootstrapProviderReady, err := waitForObjectToBeDeleted(k8sclient, ctx, key, deployment)
			if err != nil {
				return false
			}
			return isBootstrapProviderReady
```

test/e2e/minimal_configuration.go
```diff
@@ -251,5 +278,146 @@ var _ = Describe("Create providers with minimal specified configuration", func()
...
+		By("Waiting for the core provider deployment to be deleted")
+		Eventually(func() bool {
+			deployment := &appsv1.Deployment{}
+			key := client.ObjectKey{Namespace: operatorNamespace, Name: coreProviderDeploymentName}
+			isReady, err := waitForObjectToBeDeleted(k8sclient, ctx, key, deployment)
+			if err != nil {
+				return false
+			}
+			return isReady
+		}, timeout).Should(Equal(true))
```
furkatgofurov7, line 421: perhaps it is better if we will move this code to the end of core provider tests at the beginning of the file and drop the dedicated "It" test case to shorten the code.

furkatgofurov7 (commented):
Few suggestions inline, overall looks good to me

test/e2e/minimal_configuration.go
Fedosin: Yes, I wrote a test for that (line 396). The thing is that in order to install other providers, we need Core Provider to be Ready. It means we can't delete it immediately and have to postpone this action.

test/e2e/minimal_configuration.go
Fedosin: Unfortunately we can't delete it earlier, because other providers will fail a preflight check that tests that Core Provider is Ready: https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/internal/controller/preflight_checks.go#L130-L131

Fedosin:
@furkatgofurov7 thanks for your review! I updated the tests as you suggested.

test/e2e/minimal_configuration.go
furkatgofurov7: Ah right, was thinking the same initially, thanks for clarification

furkatgofurov7: @Fedosin feel free to _resolve_ the conversation

furkatgofurov7:
> @furkatgofurov7 thanks for your review! I updated the tests as you suggested.

@Fedosin thanks, looks good to me. The only thing might worth mentioning is, we could refactor the code around `waiting for provider deployment/provider` itself to be ready/deleted, check for `status.IntalledVersion` and make it generic so that we could reuse it by just calling instead. That being said, it can be a follow-up PR once this is merged 🙂

furkatgofurov7 (commented):
/lgtm

Fedosin:
@furkatgofurov7 ack, agree with you!

alexander-demicev (approved):
/approve

PR #126, opened 2023-06-01
🌱 Add go.work/go.sum to .gitignore

furkatgofurov7:
Signed-off by: Furkat Gofurov <furkat.gofurov@suse.com>

VSCode doesn't support nested modules and locally I am seeing few issues with it, i.e it will not give autocomplete or linting suggestions. Adding a go.work/go.sum files allows VSCode to recognize the nested modules.


Fedosin (approved):
/lgtm

PR #131, opened 2023-06-04
🐛 Download and store manifests from provider URL

### LINKED ISSUE #119: Cluster API Operator keeps refetching data from the specified URL after successful provider installation

Deploy a provider either from a predefined location or from a custom URL, make sure it is ready. For example:
```
kind: CoreProvider
metadata:
  name: cluster-api
  namespace: capi-system
spec:
  version: v1.4.2
```

During each subsequent reconciliation (every 10 minutes by default), the operator will repeatedly fetch the configuration from GitHub. If a network issue arises or GitHub is unavailable, the operator will report an error and mark the provider as Unhealthy, even though it is functioning properly.

I would like the CAPI operator to retrieve the provider configuration solely during the initial installation. The status of my provider should not be impacted by any GitHub-related issues.

**Environment:**

- Cluster-api version: 1.4.2
- Minikube/KIND version: 0.26.2
- Kubernetes version: (use `kubectl version`): 1.26

/kind bug

Fedosin:
Currently Cluster API Operator keeps refetching data from the specified URL after successful provider installation. It means that in case of network error Operator can't download manifests again and marks providers as unhealthy.

To prevent this, we fetch manifests only once in the beginning and store them in a configmap. For all the following reconciliations we take data from the configmap instead of downloading it from github/gitlab.

Fixes #119

### DIFF
--- a/internal/controller/manifests_downloader.go
+++ b/internal/controller/manifests_downloader.go
@@ -24,8 +24,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	operatorv1 "sigs.k8s.io/cluster-api-operator/api/v1alpha1"
-	"sigs.k8s.io/cluster-api-operator/util"
-	configclient "sigs.k8s.io/cluster-api/cmd/clusterctl/client/config"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -70,25 +68,6 @@ func (p *phaseReconciler) downloadManifests(ctx context.Context) (reconcile.Resu
 		return reconcile.Result{}, nil
 	}
 
-	// Load provider's secret and config url.
-	reader, err := p.secretReader(ctx)
-	if err != nil {
-		return reconcile.Result{}, wrapPhaseError(err, "failed to load the secret reader", operatorv1.PreflightCheckCondition)
-	}
-
-	// Initialize a client for interacting with the clusterctl configuration.
-	p.configClient, err = configclient.New("", configclient.InjectReader(reader))
-	if err != nil {
-		return reconcile.Result{}, err
-	}
-
-	// Get returns the configuration for the provider with a given name/type.
-	// This is done using clusterctl internal API types.
-	p.providerConfig, err = p.configClient.Providers().Get(p.provider.GetName(), util.ClusterctlProviderType(p.provider))
-	if err != nil {
-		return reconcile.Result{}, wrapPhaseError(err, operatorv1.UnknownProviderReason, operatorv1.PreflightCheckCondition)
-	}
-
 	repo, err := repositoryFactory(p.providerConfig, p.configClient.Variables())
 	if err != nil {
 		err = fmt.Errorf("failed to create repo from provider url for provider %q: %w", p.provider.GetName(), err)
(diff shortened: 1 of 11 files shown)


alexander-demicev (approved):
/approve

internal/controller/manifests_downloader.go
```diff
@@ -0,0 +1,154 @@
...
+		return reconcile.Result{}, wrapPhaseError(err, "failed to check that config map with manifests exists", operatorv1.PreflightCheckCondition)
+	}
+
+	if exists {
+		return reconcile.Result{}, nil
+	}
+
+	log := ctrl.LoggerFrom(ctx)
+
+	log.Info("Downloading provider manifests")
```
furkatgofurov7, line 65: we could add a verbosity to the logger:
```suggestion
	log.V(5).Info("Downloading provider manifests")
```

internal/controller/manifests_downloader.go
```diff
@@ -0,0 +1,154 @@
...
+	metadataFile, err := repo.GetFile(p.options.Version, metadataFile)
+	if err != nil {
+		err = fmt.Errorf("failed to read %q from the repository for provider %q: %w", metadataFile, p.provider.GetName(), err)
+
+		return reconcile.Result{}, wrapPhaseError(err, operatorv1.ComponentsFetchErrorReason, operatorv1.PreflightCheckCondition)
+	}
+
+	componentsFile, err := repo.GetFile(p.options.Version, repo.ComponentsPath())
+	if err != nil {
+		err = fmt.Errorf("failed to read %q from provider's repository %q: %w", repo.ComponentsPath(), p.providerConfig.ManifestLabel(), err)
```
furkatgofurov7, line 84:
```suggestion
		err = fmt.Errorf("failed to read %q from the repository for provider %q: %w", componentsFile, p.provider.GetName(), err)
```

internal/controller/manifests_downloader.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	var configMapList corev1.ConfigMapList
+
+	if err := p.ctrlClient.List(ctx, &configMapList, listOpts...); err != nil {
+		return false, fmt.Errorf("failed to list ConfigMaps: %w", err)
+	}
+
+	if len(configMapList.Items) > 1 {
+		return false, fmt.Errorf("more than one config map were found for given selector: %v", labelSelector.String())
```
furkatgofurov7, line 112:
```suggestion
		return false, fmt.Errorf("more than one config maps were found for given selector: %v", labelSelector.String())
```

internal/controller/phases.go
```diff
@@ -136,16 +132,36 @@ func (p *phaseReconciler) load(ctx context.Context) (reconcile.Result, error) {
...
-	} else {
-		p.repo, err = repositoryFactory(p.providerConfig, p.configClient.Variables())
+	return reconcile.Result{}, nil
+}
+
+// load provider specific configuration into phaseReconciler object.
+func (p *phaseReconciler) load(ctx context.Context) (reconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	log.Info("Loading provider")
```
furkatgofurov7, line 149:
```suggestion
	log.V(5).Info("Loading provider")
```

furkatgofurov7 (commented):
Few nits, overall LGTM, thanks for the PR @Fedosin

internal/controller/phases.go
alexander-demicev: We actually moved these logs to the 0 level some time ago so the operator is not too silent by default

furkatgofurov7: @alexander-demicev you mean in all cases? Because I spotted https://github.com/kubernetes-sigs/cluster-api-operator/blob/f51695b89bd11389418d9a8d45f53f88be133f70/internal/controller/phases.go#L142 where it is using the log level, but if that is just a left out then you can disregard my comment

Fedosin: I think setting log level is not strictly related to this PR, so for now we can keep it as-is and address it later.

furkatgofurov7 (commented):
/lgtm

PR #137, opened 2023-06-06
🐛 Update linters and fix reported issues

Fedosin:

This PR enables additional linters and removes unused (commented) ones. All issues reported by linters are resolved either with "make lint-fix" or manually.

Noticeably, we fix implicit memory aliasing in for loops. In golang the iteration variable is reused. At each iteration, the value of the next element in the range expression is assigned to the iteration variable; "provider" doesn't change, only its value changes. Hence, the expression &provider is referring to the same location in memory.

The recommended solution for the issue is to redeclare "provider" inside the for loop.

For more information: https://github.com/golang/go/discussions/56010

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -19,32 +19,29 @@ linters:
     - typecheck
     - unused
     # Additional linters
+    - asasalint
     - asciicheck
     - bidichk
     - bodyclose
     - contextcheck
-    # - cyclop
     - dogsled
-    # - dupl
     - durationcheck
     - errname
-    # - errorlint
+    - errorlint
     - exhaustive
-    # - exportloopref
+    - exportloopref
     - forcetypeassert
-    # - funlen
-    # - gochecknoglobals
-    # - gocognit
+    - ginkgolinter
     - goconst
     - gocritic
     - gocyclo
     - godot
-    # - goerr113
     - gofmt
+    - gofumpt
     - goheader
     - goimports
     - goprintffuncname
-    # - gosec
+    - gosec
     - importas
     - makezero
     - misspell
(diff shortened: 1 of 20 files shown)


.golangci.yaml
```diff
@@ -141,8 +147,9 @@ issues:
     # Exclude some linters from running on tests files.
     - path: _test\.go
       linters:
-        - gocyclo
-        - dupl
         - gosec
-        - gochecknoglobals
-        - goerr113
+    - path: envtest/environment.go
```
furkatgofurov7, line 151: where is this located?

cmd/main.go
```diff
@@ -130,7 +128,12 @@ func main() {
 		klog.Infof("Profiler listening for requests at %s", profilerAddress)
 
 		go func() {
-			klog.Info(http.ListenAndServe(profilerAddress, nil))
+			server := &http.Server{
+				Addr:              profilerAddress,
+				ReadHeaderTimeout: 3 * time.Second,
+			}
```
furkatgofurov7, line 134: interesting, was this also reported as issue in linter? How are we coming up with 3 seconds for `ReadHeaderTimeout`

furkatgofurov7 (commented):
Thanks for improvement, overall looks good, 2 questions inline:

.golangci.yaml
alexander-demicev: it's now in `internal` package

furkatgofurov7: ah yes so it has to be: `internal/envtest/environment.go`

Fedosin: fixed :)

cmd/main.go
Fedosin: yep, it was reported by `gosec` linter. I thought that 3 second timeout is sane...

alexander-demicev (approved):
/approve

Issue #130, opened 2023-06-01
Helm chart support more comprehensive installation


dtzar:
**User Story**

As an operator I would like to have a single helm chart to install not only the operator, but the entire e2e stack to have a functioning management cluster with the provider installed so that it is quick and easy to maintain and upgrade.

**Detailed Description**
There are many disconnected steps in the quick start guide and the helm chart should be able to support any of the providers versus manually applying various independent yaml files.

/kind feature

alexander-demicev:
Already working on it
/assign

PR #144, opened 2023-06-07
✨ Add cert-manager as a dependency to helm-chart

Fedosin:

To work properly CAPI Operator needs cert-manager. To ensure it is installed, we add a cert-manager dependency to the helm chart.

Partially-Fixes: https://github.com/kubernetes-sigs/cluster-api-operator/issues/130


Fedosin:
Blocked by https://github.com/kubernetes-sigs/cluster-api-operator/pull/145

damdo:
How does this behave when the cluster already has cert manager installed in a different way and not through Helm?
How does this behave when the cluster already has cert manager installed via Helm but that's a different version from the one linked here as a dependency?

Fedosin:
Yeah, I think this change just over-complicates the helm chart. Moreover, they don't support installing CRDs and some manual actions are required anyway.
I'm going to close this PR.

Fedosin:
Seems working now. I tried it with `helm install cluster-api-operator out/package/cluster-api-operator-0.3.0.tgz --set cert-manager.enabled=true --create-namespace -n capi-operator-system --set cert-manager.namespace=cert-manage`
Everything is installed successfully in related namespaces.

alexander-demicev:
@Fedosin is there an option to disable cert-manager installation?

g-gaston:
/lgtm
thanks a lot for doing this!

Fedosin:
@alexander-demicev Helm installation is disabled by default: https://github.com/kubernetes-sigs/cluster-api-operator/pull/144/files#diff-eb3428ea28fccc0ad2deffddc87823203d05dc4a3b5e9d73f3f093ecc40f5a53R3

You need to explicitly set `cert-manager.enabled=true` to install it. I provided an example of helm command that does it: https://github.com/kubernetes-sigs/cluster-api-operator/pull/144#issuecomment-1605185676

alexander-demicev:
I'm okay with merging this if others have no objections @damdo @furkatgofurov7 we should also update docs in a follow-up PR

alexander-demicev:
We can also download cert-manager chart during `make release` instead of keeping it in repo

Fedosin:
@alexander-demicev yeah, it makes sense. I also did it for `cert-manager-v1.12.2.tgz` binary, so we won't store it in our repo.

furkatgofurov7 (commented):
/lgtm

PR #145, opened 2023-06-07
🐛 prevent false lint error

Fedosin:

Gosec linter reports a false error `G101: Potential hardcoded credentials (gosec)`. To prevent it we disable the linter in that place.


damdo (commented):
/approve

furkatgofurov7 (commented):
/lgtm

PR #151, opened 2023-06-08
🌱 Include chart-update folder to verify-modules Makefile target

furkatgofurov7:
Includes `chart-update` folder,  which includes go.mod/go.sum files, to `verify-modules` Makefile target so that we do not miss go mod files and update all files at once

Fixes #


Fedosin (approved):
/lgtm

PR #154, opened 2023-06-11
📖 Add operator documentation

Fedosin:

Currently there is not so much documentation written. The proposal doesn't count, and also it is outdated and sometimes misleading. This PR fixes this issue and adds full documentation for the operator.


alexander-demicev (approved):
/approve

Thanks for cleaning up and adding new docs, it will be extremely useful!

furkatgofurov7 (commented):
This is very comprehensive and detailed documentation, thanks @Fedosin working on this!

I have two questions:

1. I am not sure if I am in favour of maintaining two docs (README.md vs original operator proposal) which are very similar to each other, because you need to update both when we will do a minor/breaking changes in the future. 

Also, it makes sense to have the original operator proposal in the repo, maybe it is better not to touch it at all and make somehow clear that it is the early version of operator (since it is a proposal at the end) and someone who needs the latest doc, refer to README.md introduced in this PR. 

2. Perhaps we can get rid of the https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/docs/getting-started.md completely since it is already covered?

Fedosin:
@furkatgofurov7 agree... plus we need to update https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/README.md accordingly

docs/README.md
```diff
@@ -0,0 +1,640 @@
...
+The first step is to install the CoreProvider, which is responsible for managing the Cluster API CRDs and the Cluster API controller.
+
+*Example:*
+
+```yaml
+apiVersion: operator.cluster.x-k8s.io/v1alpha1
+kind: CoreProvider
+metadata:
+  name: cluster-api
+  namespace: capi-system
```
furkatgofurov7, line 153: this ns has to be already created otherwise it would fail, can we mention/append it somewhere?

Fedosin: Thank you for review! I addressed that part

furkatgofurov7 (commented):
This version looks a lot better, thanks @Fedosin, we can re-iterate later if we find out something is missing.

/lgtm

PR #162, opened 2023-06-14
🐛 use correct conditions for reporting errors after preflight checks

Fedosin:

Sometimes we use PreflightCheckCondition to report errors when preflight validations already passed, which is not correct. We need to use ProviderInstalledCondition in such cases.

Fixes #


alexander-demicev (approved):
/approve

furkatgofurov7 (commented):
/lgtm

PR #166, opened 2023-06-16
🌱 Relocate version change validation to preflight check 

Fedosin:

Previously, the check was executed at the end of reconciliation and in multiple instances. By moving it to the beginning, the process is halted when installed and desired versions match, streamlining the reconciliation workflow.

### DIFF
--- a/internal/controller/genericprovider_controller.go
+++ b/internal/controller/genericprovider_controller.go
@@ -108,10 +108,15 @@ func (r *GenericProviderReconciler) Reconcile(ctx context.Context, req reconcile
 	}
 
 	// Run preflight checks to ensure that we can proceed with given specification
-	res, err := preflightchecks.PreflightChecks(ctx, r.Client, typedProvider, typedProviderList)
+	res, halt, err := preflightchecks.PreflightChecks(ctx, r.Client, typedProvider, typedProviderList)
 	if !res.IsZero() || err != nil {
-		// the steps are sequential, so we must be complete before progressing.
 		return res, err
+	}
+
+	if halt {
+		log.Info("No changes detected, skipping further steps")
+
+		return res, nil
 	}
 
 	return r.reconcile(ctx, typedProvider, typedProviderList)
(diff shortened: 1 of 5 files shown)


Fedosin:
The issue is addressed in https://github.com/kubernetes-sigs/cluster-api-operator/pull/168, so the only meaningful thing we can take from this PR is to move preflight checks into a separate package.

internal/controller/genericprovider_controller.go
```diff
@@ -127,7 +140,6 @@ func patchProvider(ctx context.Context, provider genericprovider.GenericProvider
 func (r *GenericProviderReconciler) reconcile(ctx context.Context, provider genericprovider.GenericProvider, genericProviderList genericprovider.GenericProviderList) (ctrl.Result, error) {
 	reconciler := newPhaseReconciler(*r, provider, genericProviderList)
 	phases := []reconcilePhaseFn{
-		reconciler.preflightChecks,
```
damdo, line 130: What if we kept the preflightChecks as a phase here so we can benefit from the centralised condition setting logic, and rather we defined the `halt` pattern you came up with, as a pattern for all the phases.

Maybe with a customisable halting message (similar to the one you specified [here](https://github.dev/Fedosin/cluster-api-operator/blob/b17b9c91be17e774101047058762226f192bf96c/internal/controller/genericprovider_controller.go#L117-L118)) that can be logged before returning reconcile()? We could even have `halt` as a message string ()`haltMsg`?), and when the phase returns a non empty msg then we log and halt?

So [here](https://github.com/Fedosin/cluster-api-operator/blob/5c223b751ebc4c19fc4bdcc0ba8027ecbbd56fdb/internal/controller/genericprovider_controller.go) doing something like
```
res, halt, err = phase(ctx)
...
```

internal/controller/preflightchecks/preflight_checks.go
```diff
@@ -238,3 +252,26 @@ func isPredefinedProvider(providerName string, providerType clusterctlv1.Provide
...
+	res, err := currentVersion.Compare(provider.GetSpec().Version)
+	if err != nil {
+		return false, err
+	}
+
+	// we need to delete installed components if versions are different
+	versionChanged := res != 0
+
+	return versionChanged, nil
+}
```
damdo, line 277: I know this is just codemotion, but maybe we could make this even more useful and if
`currentVersion.Compare != 0` perform a [deep.Equal()](https://pkg.go.dev/github.com/go-test/deep#Equal) and return the `[]string` with the diff between the two or else return `[]string{}, nil`.
This way we could log what is the detected difference and provide useful info to the user about it.

PR #168, opened 2023-06-18
🐛 Update provider components if (and only if) its spec has been changed

### LINKED ISSUE #156: Operator doesn't reconcile providers if their spec values changed
Deploy a provider specifying some deployment arguments. For instance:
```yaml
kind: InfrastructureProvider
metadata:
 name: aws
 namespace: capa-system
spec:
 version: v2.1.4
 secretName: aws-variables
 deployment:
   containers:
   - name: manager
     image:
         repository: "gcr.io/myregistry"
         name: "capa-controller"
         tag: "v2.1.4-foo"
```

Make sure that capa started successfully.

Update image in the provider spec:
```yaml
kind: InfrastructureProvider
metadata:
 name: aws
 namespace: capa-system
spec:
 version: v2.1.4
 secretName: aws-variables
 deployment:
   containers:
   - name: manager
     image:
         repository: "gcr.io/myregistry"
         name: "capa-controller"
         tag: "v2.1.4-bar"
```

Operator should reconcile capa provider and update image for the deployment, but nothing happens.

**Environment:**

/kind bug

Fedosin:
Now we update provider components only if it's version has changed, but there are other cases where we need to do the same. For instance, when we update its DeploymentSpec or ManagementSpec.

Also we should reconcile providers if their spec is the same as for the last applied version. 

To fix it we calculate last applied spec hash and store it in a provider annotation. For next reconciliations we calculate the hash again. If it's the same as in the annotation, we skip further steps and stop immediately.

Fixes #156

### DIFF
--- a/internal/controller/genericprovider_controller.go
+++ b/internal/controller/genericprovider_controller.go
@@ -18,6 +18,8 @@ package controller
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -43,6 +45,10 @@ type GenericProviderReconciler struct {
 	Client       client.Client
 	Config       *rest.Config
 }
+
+const (
+	appliedSpecHashAnnotation = "operator.cluster.x-k8s.io/applied-spec-hash"
+)
 
 func (r *GenericProviderReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
 	return ctrl.NewControllerManagedBy(mgr).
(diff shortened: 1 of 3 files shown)


internal/controller/genericprovider_controller.go
```diff
@@ -106,7 +112,35 @@ func (r *GenericProviderReconciler) Reconcile(ctx context.Context, req reconcile
...
 	}
 
-	return r.reconcile(ctx, typedProvider, typedProviderList)
+	// Check if spec hash stays the same and don't go further in this case.
+	specHash, err := calculateHash(typedProvider.GetSpec())
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if typedProvider.GetAnnotations()[appliedSpecHashAnnotation] == specHash {
```
alexander-demicev, line 121: If I remember it right, `GetAnnotations()` can still return nil

internal/controller/genericprovider_controller.go
```diff
@@ -227,3 +261,19 @@ func (r *GenericProviderReconciler) newGenericProviderList() (genericprovider.Ge
...
+
+func calculateHash(object interface{}) (string, error) {
+	jsonData, err := json.Marshal(object)
+	if err != nil {
+		return "", fmt.Errorf("cannot parse provider spec: %w", err)
+	}
+
+	hash := sha256.New()
+
+	_, err = hash.Write(jsonData)
```
alexander-demicev, line 273:
```suggestion
	if _, err := hash.Write(jsonData); err != nil {
```

internal/controller/genericprovider_controller.go
Fedosin: Yeah, it can, but it doesn't matter. We can get values from nil maps. Here is an example I wrote: https://go.dev/play/p/pUL9u7U5OyR

alexander-demicev (approved):
/approve

cc @damdo @furkatgofurov7

furkatgofurov7 (commented):
/lgtm

PR #169, opened 2023-06-18
🐛 Delete provider's CRDs and namespace along with the provider custom resource

### LINKED ISSUE #120: CRDs are not deleted when provider is removed
1. Install a provider:
```
kind: CoreProvider
metadata:
  name: cluster-api
  namespace: capi-system
spec:
  version: v1.4.2
```
2. Delete the provider
3. Ensure that all previously created CRDs from `cluster.x-k8s.io` are not deleted together with the provider.

All created resources should be removed.

**Environment:**

- Cluster-api version: 1.4.2
- Minikube/KIND version: 1.27
- Kubernetes version: (use `kubectl version`): 1.27

/kind bug

Fedosin:

Now we keep provider's namespace and CRDs after we delete the provider, which pollutes the system. System administrators have to manually delete stalled objects after that. After this PR operator will delete all providers components completely.

Fixes #120

### DIFF
--- a/internal/controller/client_proxy.go
+++ b/internal/controller/client_proxy.go
@@ -76,6 +76,8 @@ func (k *controllerProxy) ListResources(labels map[string]string, namespaces ...
 				{Kind: "Secret", Namespaced: true},
 				{Kind: "ConfigMap", Namespaced: true},
 				{Kind: "Service", Namespaced: true},
+				{Kind: "ServiceAccount", Namespaced: true},
+				{Kind: "Namespace"},
 			},
 		},
 		{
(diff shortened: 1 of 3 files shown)


Fedosin:
Maybe it makes sense to split this PR in two parts: one related to CRDs and namespaces, and another to all other cluster-scoped resources. For instance, now we don't delete clusterroles and clusterrolebinding, which is clearly a bug.

But ideally, we need to discuss the process of provider deletion in detail. Now namespaced resources are deleted because they have an owner reference to the provider object, and some cluster scoped resources are removed with clusterctl's delete. I think it's inconsistent and should be improved.

furkatgofurov7:
> The initial idea was to have an identical behavior as `clusterctl` https://cluster-api.sigs.k8s.io/clusterctl/commands/delete.html#clusterctl-delete where namespaces and CRDs are not deleted. cc @richardcase @furkatgofurov7

Agree. Thinking of a use case we can of course delete everything in the most common use case we have that is single-tenant management clusters. However I think it might be destructive in multi-tenant scenario, because then we would delete both instance specific and shared CRDs/web-hook components.

Fedosin:
/hold

We decided to split this PR in two parts - [one](https://github.com/kubernetes-sigs/cluster-api-operator/pull/179) that fixes the bug with undeleted resources, and another that allows to regulate whether namespace and CRDs should be deleted or not.

Fedosin:
Close it for now...

PR #173, opened 2023-06-19
🐛 Always call customizeDeployment function

### LINKED ISSUE #172: Feature gates are not injected into manager pod

When I want to set custom feature gates for a provider manager, they are ignored

Example:

```yaml
kind: CoreProvider
metadata:
  name: cluster-api
spec:
  version: v1.4.3
  manager:
    featureGates: 
      MachinePool: true
      ClusterResourceSet: true
      ClusterTopology: true
      RuntimeSDK: false
      LazyRestmapper: false    
```

For this manifests operator doesn't modify feature gates, and leaves default values, i.e. `--feature-gates=MachinePool=false,ClusterResourceSet=false,ClusterTopology=false,RuntimeSDK=false,LazyRestmapper=false`

Feature gates should be updated like `--feature-gates=MachinePool=true,ClusterResourceSet=true,ClusterTopology=true,RuntimeSDK=false,LazyRestmapper=false`

**Environment:**

- Cluster-api-operator version:
- Cluster-api version:
- Minikube/KIND version:
- Kubernetes version: (use `kubectl version`):

/kind bug

Fedosin:
Now we call it only if Deployment spec is not empty. Then, inside the function, we customize both Deployment and Manager if related specs are not empty. It causes an error when we don't update Manager if Deployment spec is empty.

Fixes #172

### DIFF
--- a/internal/controller/component_customizer.go
+++ b/internal/controller/component_customizer.go
@@ -73,9 +73,7 @@ func customizeObjectsFn(provider genericprovider.GenericProvider) func(objs []un
 					return nil, err
 				}
 
-				if provider.GetSpec().Deployment != nil {
-					customizeDeployment(provider.GetSpec(), d)
-				}
+				customizeDeployment(provider.GetSpec(), d)
 
 				if err := scheme.Scheme.Convert(d, &o, nil); err != nil {
 					return nil, err


damdo (commented):
/lgtm

PR #179, opened 2023-06-22
🐛 delete all provider resources 

Fedosin:

Client proxy defines among which types we will search for objects we need to delete. Currently this list is not full and we skip some types, like ClusterRole or ClusterRole bindings. It leads to the fact that objects of these types stay in the system when their provider is deleted.

This PR updates the list of resource types we need to scan.

Fixes #


damdo (commented):
/lgtm

Fedosin:
```
docker pull docker.io/library/golang:1.19.0
Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
make: *** [Makefile:318: docker-pull-prerequisites] Error 1
```
>_<

PR #180, opened 2023-06-25
✨Fetch latest provider version if it's not set in the spec

Fedosin:

Now users have to manually set provider version, which is not convenient as it requires additional github searching. This PR allows to omit it, and the latest version will be substituted automatically. This behavior is similar to what is done in clusterctl, where version is optional.

Additionally this PR adds e2e tests for air-gapped environments.

Fixes #


internal/controller/phases.go
```diff
@@ -143,6 +138,83 @@ func (p *phaseReconciler) initializePhaseReconciler(ctx context.Context) (reconc
...
+	log := ctrl.LoggerFrom(ctx)
+
+	log.Info("Setting the latest provider version.")
+
+	var repo repository.Repository
+
+	var err error
+
+	// Create repository either from provided URL or from a config map.
+	if p.provider.GetSpec().FetchConfig != nil && p.provider.GetSpec().FetchConfig.Selector != nil {
```
alexander-demicev, line 158: can we optimize it and initialize the repo only once? It gets created in the later phases too

Fedosin:
/hold 

@alexander-demicev agree, we can optimise this code a little. I'll propose another update for this PR soon

api/v1alpha1/provider_types.go
```diff
@@ -31,7 +31,7 @@ const (
 // ProviderSpec is the desired state of the Provider.
 type ProviderSpec struct {
 	// Version indicates the provider version.
```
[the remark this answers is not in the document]
Fedosin: fixed

internal/controller/phases.go
Fedosin: yeah, I slightly rewrote the original PR, so it is more optimized now.

test/e2e/air_gapped_test.go
```diff
@@ -0,0 +1,282 @@
...
+	customManifestsFolder = "resources/"
+)
+
+var _ = Describe("Install Core Provider in an air-gapped environment", func() {
+	It("should successfully create config maps with Core Provider manifests", func() {
+		k8sclient := bootstrapClusterProxy.GetClient()
+
+		configMaps := []corev1.ConfigMap{}
+
+		for _, fileName := range []string{"core-cluster-api-v1.4.2.yaml", "core-cluster-api-v1.4.3.yaml"} {
```
furkatgofurov7, line 45: perhaps this the only way to do it in air gapped env (manually updating/adding huge manifests as CAPI releasing new versions), but I do not have a better suggestion either how we could improve it..

furkatgofurov7 (commented):
What happens in case, CAPI tags v1.5.x as latest while CAPZ did not yet uplift to latest of CAPI and has latest tagged version compatible only with CAPI v1.4.x?

alexander-demicev:
@furkatgofurov7 I guess the issue here is similar to when using `clusterctl`, you have to specify version in case you want to be sure that dependencies are aligned

test/e2e/air_gapped_test.go
alexander-demicev: yes, this is the only way. For `clusterctl` you have to create local repositories, for operator you can download manifests from github and make a configmap from file.

furkatgofurov7:
> @furkatgofurov7 I guess the issue here is similar to when using `clusterctl`, you have to specify version in case you want to be sure that dependencies are aligned

That is true and I guess it is fine to follow the same behaviour here

alexander-demicev (approved):
/approve

furkatgofurov7:
/hold
to address https://github.com/kubernetes-sigs/cluster-api-operator/pull/180#discussion_r1252021483

internal/controller/manifests_downloader.go
```diff
@@ -75,15 +76,25 @@ func (p *phaseReconciler) downloadManifests(ctx context.Context) (reconcile.Resu
...
 	}
 
+	spec := p.provider.GetSpec()
+
+	if spec.Version == "" {
+		// User didn't set the version, try to get repository default.
+		spec.Version = repo.DefaultVersion()
+
+		// Add version to the provider spec.
+		p.provider.SetSpec(spec)
```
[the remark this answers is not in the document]
Fedosin: The reason spec is set twice is because it accommodates two distinct workflows:

- When downloading manifests from GitHub/GitLab.
- When working with config maps containing manifests.

For the first workflow, the version is set during the manifest download process when a fetch URL is used. In the case of config maps, the version setting occurs in the load phase. Regardless of the workflow, the spec update is executed only once.

PR #190, opened 2023-07-14
✨ Introduce v1alpha2 API

### LINKED ISSUE #167: v1alpha2?
Let's discuss ideas that require a change in the API version.

1. [ImageMeta](https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/api/v1alpha1/provider_types.go#L151-L164) can be presented as an image url, i.e. <repository>/<image_name>:<tag> instead of the current object.
2. [Version](https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/api/v1alpha1/provider_types.go#L34) in ProviderSpec is now mandatory, but technically we can omit it and substitute the latest available provider version at the time of installation.

/kind feature

Fedosin:
This PR introduces new API version `v1alpha2`.

Fixes #167


test/e2e/helm_test.go
```diff
@@ -39,10 +39,12 @@ var _ = Describe("Create a proper set of manifests when using helm charts", func
 			AdditionalFlags: []string{"-n=capi-operator-system", "--create-namespace"},
 		}
 		fullInstallChart.Output = Manifests
-		manifests, err := fullInstallChart.InstallChart(nil)
+		_, err := fullInstallChart.InstallChart(nil)
 		Expect(err).ToNot(HaveOccurred())
-		fullChartInstall, err := os.ReadFile(filepath.Join(customManifestsFolder, "full-chart-install.yaml"))
-		Expect(manifests).To(Equal(string(fullChartInstall)))
+
+		// TODO(mfedosin): we need to rethink this check because any change to the chart causes it to fail.
```
Danil-Grigorev, line 45: That is the goal. It will be visible what is changed each time a PR is up. Please update it so we can see the diff

Danil-Grigorev: I’ve added https://github.com/kubernetes-sigs/cluster-api-operator/pull/212 to generate the expected test output.

.golangci.yaml
```diff
@@ -133,7 +133,7 @@ linters-settings:
       - pkg: sigs.k8s.io/cluster-api/cmd/clusterctl/client/config
         alias: configclient
       # CAPI Operator
-      - pkg: sigs.k8s.io/cluster-api-operator/api/v1alpha1
+      - pkg: sigs.k8s.io/cluster-api-operator/api/v1alpha2
```
furkatgofurov7, line 136: I do not think we need to remove this, but rather add a new entry for new API version because you are using current API version in https://github.com/kubernetes-sigs/cluster-api-operator/pull/190/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R43 as `operatorv1alpha1`?

```suggestion
      - pkg: sigs.k8s.io/cluster-api-operator/api/v1alpha1
        alias: operatorv1alpha1
```

internal/controller/component_customizer.go
```diff
@@ -257,8 +257,8 @@ func customizeContainer(cSpec operatorv1.ContainerSpec, d *appsv1.Deployment) {
 				c.Resources = *cSpec.Resources
 			}
 
-			if cSpec.Image != nil && cSpec.Image.Name != "" && cSpec.Image.Repository != "" {
-				c.Image = imageMetaToURL(cSpec.Image)
```
furkatgofurov7, line 261: how are these changes related to new API version?

furkatgofurov7 (commented):
Thank you working on this, few suggestions inline:

furkatgofurov7:
Also forgot to mention, I would consider this as a breaking change and suggest re-titling the PR (with ⚠️) to signal it properly in the release notes

internal/controller/component_customizer.go
Danil-Grigorev: I guess this is the reason for the new API version. Incompatible fields

Fedosin: Yes, @Danil-Grigorev is right. It's a result of the API change. In v1alpha1 we used ImageMeta object to provide a custom image for the provider. In v1alpha2 we decided that it's better to use the well-known <registry>/<name>:<tag> image URL format, which is just a string.
After this change we don't have to convert ImageMeta -> ImageURL anymore, hence the change.

Danil-Grigorev: Should we add documentation for user migration from v1alpha1 to v1alpha2 in this case? Similar to https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/providers/version-migration.md?plain=1

furkatgofurov7: But migration doc in CAPI is formed when they release a minor release, and describe changes needed going from one minor release of CAPI to another.

Fedosin: Technically, you don't need to migrate anything manually. I implemented auto-conversion, which handles v1alpha1 -> v1alpha2 upgrade.
But I agree that it would be nice to have a doc describing the changes in the new version.

Danil-Grigorev (approved):
/lgtm

Danil-Grigorev:
Thanks @Fedosin, looking good!

furkatgofurov7 (approved):
/approve
/hold


1. pending clarification on: https://github.com/kubernetes-sigs/cluster-api-operator/pull/190#discussion_r1294982521. I would prefer taking care of this with the same PR if we decide to go down that path but no strong opinions since we have a autoconversion in place.
2. in case we want to have more eyes on it

Fedosin:
@furkatgofurov7 I agree with you - some documentation is necessary here. I'll write it later today and update the PR.

docs/README.md
```diff
@@ -435,6 +425,21 @@ The following sections provide details about `ProviderSpec` and `ProviderStatus`
          matchLabels:
    ...
    ```
+
+6. `SecretReference`: pointer to a secret object, , consisting of:
```
furkatgofurov7, line 429: double `,` in the sentence

docs/migrating.md
```diff
@@ -0,0 +1,57 @@
+# Migrating from v1alpha1 to v1alpha2
+
+This document describes changes that were introduced in v1alpha2 API and how to update your templates to the new version.
+
+## ImageMeta -> imageURL conversion
```
furkatgofurov7, line 5: One more suggestion: we could have headers in the doc, similar to https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/providers/migrations/v1.4-to-v1.5.md#cluster-api-v14-compared-to-v15 and for example for API changes like below it could be https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/providers/migrations/v1.4-to-v1.5.md#changes-by-kind

docs/migrating.md
```diff
@@ -0,0 +1,57 @@
+# Migrating from v1alpha1 to v1alpha2
```
furkatgofurov7, line 1: How about having a separate folder for migrations doc instead, i.e `docs/migrations/v1alpha1-to-v1alpha2.md` so that we keep it separate and add new migration docs in the future following the above format?

furkatgofurov7:
@Fedosin thanks a lot, this version looks great to me!

furkatgofurov7 (commented):
/lgtm

Pending CI to pass (currently failing due to API rate limit and needs re-run after an hour)

Danil-Grigorev (commented):
/lgtm

PR #192, opened 2023-07-19
🐛 Compress components manifests if they don't fit in the configmap 

### LINKED ISSUE #20: Merge content from ConfigMaps
ProviderSpec.FetchConfiguration support a label selector for providing ConfigMap as a source for provider's components https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/api/v1alpha1/provider_types.go#L186, if a provider decides to split it's components into multiple configmaps this won't work because our code doesn't support it https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/controllers/phases.go#L216. We should merge content from ConfigMaps of the same provider version into one piece of data before storing it in MemoryRepository https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/controllers/phases.go#L239.

/kind feature

Fedosin:
By default we download capi provider manifests from the provided URL and store it in a confimap to reuse later. Currently there is a limit on the configmap size, which is 1mb. If we the manifests size is bigger then the installation process fails.

To avoid it we will compress manifests if they can't be stored in the configmap directly.
Fixes #20

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -664,3 +664,42 @@ spec:
     selector:
       matchLabels:
         provider-components: azure
+```
+
+### Situation when manifests do not fit into configmap
+
+There is a limit on the [maximum size](https://kubernetes.io/docs/concepts/configuration/configmap/#motivation) of a configmap - 1MiB. If the manifests do not fit into this size, Kubernetes will generate an error and provider installation fail. To avoid this, you can archive the manifests and put them in the configmap that way.
+
+For example, you have two files: `components.yaml` and `metadata.yaml`. To create a working config map you need:
+
+1. Archive components.yaml using `gzip` cli tool
+
+```sh
+gzip -c components.yaml > components.gz
+```
+
+2. Create a configmap manifest from the archived data
+
+```sh
+kubectl create configmap v1.9.3 --namespace=capz-system --from-file=components=components.gz --from-file=metadata=metadata.yaml --dry-run=client -o yaml > configmap.yaml
+```
+
+3. Edit the file by adding "provider.cluster.x-k8s.io/compressed: true" annotation
+
+```sh
+yq eval -i '.metadata.annotations += {"provider.cluster.x-k8s.io/compressed": "true"}' configmap.yaml
+```
+
+**Note**: without this annotation operator won't be able to determine if the data is compressed or not.
+
+4. Add labels that will be used to match the configmap in `fetchConfig` section of the provider
+
+```sh
+yq eval -i '.metadata.labels += {"my-label": "label-value"}' configmap.yaml
+```
+
+5. Create a configmap in your kubernetes cluster using kubectl
+
+```sh
+kubectl create -f configmap.yaml
+```
--- a/internal/controller/manifests_downloader.go
+++ b/internal/controller/manifests_downloader.go
@@ -17,6 +17,8 @@ limitations under the License.
 package controller
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
 	"fmt"
 
@@ -36,8 +38,12 @@ const (
 	configMapNameLabel    = "provider.cluster.x-k8s.io/name"
 	operatorManagedLabel  = "managed-by.operator.cluster.x-k8s.io"
 
+	compressedAnnotation = "provider.cluster.x-k8s.io/compressed"
+
 	metadataConfigMapKey   = "metadata"
 	componentsConfigMapKey = "components"
+
+	maxConfigMapSize = 1 * 1024 * 1024
 )
 
 // downloadManifests downloads CAPI manifests from a url.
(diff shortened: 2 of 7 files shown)


internal/controller/phases.go
```diff
@@ -279,6 +282,41 @@ func (p *phaseReconciler) configmapRepository(ctx context.Context, labelSelector
 	return mr, nil
 }
 
+// getComponentsData returns components data based on if it's compressed or not.
+func getComponentsData(cm corev1.ConfigMap) (string, error) {
+	// Data is not compressed, return it immediately.
+	if cm.GetAnnotations()[compressedAnnotation] != "true" {
```
alexander-demicev, line 288: We should also add docs about this annotation, If users compress manifests manually and try to use them in an air-gapped scenario, this annotation has to be set on the configmap.

internal/controller/manifests_downloader.go
```diff
@@ -151,11 +158,34 @@ func (p *phaseReconciler) createManifestsConfigMap(ctx context.Context, metadata
 			Labels:    p.prepareConfigMapLabels(),
 		},
 		Data: map[string]string{
-			metadataConfigMapKey:   metadata,
-			componentsConfigMapKey: components,
+			metadataConfigMapKey: string(metadata),
 		},
 	}
 
+	// Components manifests data can exceed the configmap size limit. In this case we have compress it.
```
furkatgofurov7, line 165:
```suggestion
	// Components manifests data can exceed the configmap size limit. In this case we have to compress it.
```

internal/controller/phases.go
```diff
@@ -279,6 +282,41 @@ func (p *phaseReconciler) configmapRepository(ctx context.Context, labelSelector
...
+		components, ok := cm.Data[componentsConfigMapKey]
+		if !ok {
+			return "", fmt.Errorf("ConfigMap %s/%s Data has no components", cm.Namespace, cm.Name)
+		}
+
+		return components, nil
+	}
+
+	// Otherwise we have to decompress the data first.
```
furkatgofurov7, line 297: Q to better understand this logic: what is the use case when it is already compressed and why we need to decompress in that case?

furkatgofurov7 (commented):
Thank you!

Overall LGTM, small nit and not blocking question

internal/controller/phases.go
Fedosin: In the normal workflow, when we create a provider for the first time, we download manifests from the provided URL and store them in a configmap to reuse later. Unfortunately, there is a size limit (1MiB) on the configmap in kubernetes, and for some providers we can't store there manifests directly.

Here is an example of capi provider that has infrastructure-components.yaml file exceeding 1MiB: https://github.com/oracle/cluster-api-provider-oci/releases/tag/v0.12.0

The error looks like this:
```
E0718 10:19:28.002488       1 controller.go:329]  "msg"="Reconciler error" "error"="failed to create config map for provider \"oci\": ConfigMap \"infrastructure-oci-v0.12.0\" is invalid: []: Too long: must have at most 1048576 bytes" "InfrastructureProvider"={"name":"oci","namespace":"cluster-api"} "controller"="infrastructureprovider" "controllerGroup"="operator.cluster.x-k8s.io" "controllerKind"="InfrastructureProvider" "name"="oci" "namespace"="cluster-api" "reconcileID"="
3530ba70-285c-4a14-997b-5bdfc092c973
```

To workaround this we put compressed data in the configmap, which is 10 times smaller than the original. Later, when we generate clusterctl provider repository in the "load" phase, we decompress it and use as before.

internal/controller/phases.go
Fedosin: Hey! I added the documentation, that describes this process for air-gapped environments.

internal/controller/phases.go
furkatgofurov7: Thanks @Fedosin!

alexander-demicev (approved):
/approve

test/e2e/compressed_manifests_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+const (
+	ociInfrastructureProviderName           = "oci"
+	ociInfrastructureProviderVersion        = "v0.12.0"
+	ociInfrastructureProviderDeploymentName = "capoci-controller-manager"
+	compressedAnnotation                    = "provider.cluster.x-k8s.io/compressed"
+	componentsConfigMapKey                  = "components"
+)
+
+var _ = Describe("Create and delete a provider with manifests that don't fit the configmap", func() {
```
alexander-demicev, line 42: can we also have a test where we supply an already compressed configmap?

docs/README.md
```diff
@@ -664,3 +664,42 @@ spec:
...
+1. Archive components.yaml using `gzip` cli tool
+
+```sh
+gzip -c components.yaml > components.gz
+```
+
+2. Create a configmap manifest from the archived data
+
+```sh
+kubectl create configmap v1.9.3 --namespace=capz-system --from-file=components=components.gz --from-file=metadata=metadata.yaml --dry-run=client -o yaml > configmap.yaml
```
furkatgofurov7, line 684: Not a blocking one and can be tackled in a follow-up if found needed: `v1.9.3` is probably CAPZ provider version in this case, might be better to have different name (capz-cm / capz-v1.9.3 as an examples) for cm instead of just version.

furkatgofurov7 (commented):
/lgtm

PR #202, opened 2023-08-03
🐛 Support installing providers with custom names in air-gapped environments

Fedosin:
Currently we do not support installation of providers with custom names for air-gapped environments. It happens because we inject new names in the list of known providers only for those with URL specified, ignoring config maps.
This PR fixes it, so we support both URL and ConfigMap installations with custom names.

Fixes #


Danil-Grigorev (approved):
\lgtm

alexander-demicev (approved):
/approve

PR #205, opened 2023-08-03
🌱  Split out test modules from capi operator modules

Danil-Grigorev:

This small change separates capi operator modules from test suite, allowing independent use and external imports. Removes core dependency on capi testing framework. Follows similar pattern to CAPI test suite

Fixes #


Fedosin (approved):
/approve

awesome! thank you for doing this.

furkatgofurov7 (commented):
This looks good, apart couple of suggestions: 

- since we have separated now test modules into its own folder and we ship it with its own Go module, we should:
    -  add notes informing about this change properly 
    -  tag and version the new module and as part of the release. We would have to create an annotated tag for /test folder and push it every time we do a release from the repo, once this goes in. For example, if we're releasing v0.5.0, we should also tag test/v0.5.0
    - add a path to new go.mod/sum files in the Makefile [`modules`](https://github.com/kubernetes-sigs/cluster-api-operator/blob/72873f9dbe9f37a74e473a55b73fa5d49444b0a8/Makefile#L315-L319) target where we run gomod tidy in all folders where we have go.mod/sum files and [`verify-modules`](https://github.com/kubernetes-sigs/cluster-api-operator/blob/72873f9dbe9f37a74e473a55b73fa5d49444b0a8/Makefile#L273-L278) target
    - add it to Dockerfile in https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/Dockerfile#L29-L31

alexander-demicev (approved):
/lgtm
Thanks!

alexander-demicev:
I don't know if comments from @furkatgofurov7 are blocking so I put the PR on hold, feel free to remove it if needed

Danil-Grigorev:
Everything fixed. @furkatgofurov7 I think nothing is pending

docs/release.md
```diff
@@ -10,11 +10,18 @@ git push -u upstream release-0.1
 ```
 
 ```bash
+# Export the tag of the release to be cut, e.g.:
 export RELEASE_TAG=v0.1.1
 
+# Create tags locally
+# Warning: The test tag MUST NOT be an annotated tag.
 git tag -s -a ${RELEASE_TAG} -m ${RELEASE_TAG}
+git tag test/${RELEASE_TAG}
```
furkatgofurov7, line 19: let's sign this tag with gpg signature as above?

furkatgofurov7 (commented):
Last nit:

docs/release.md
Danil-Grigorev: This process is coming from CAPI release document. They explicitly don’t sign test package. I’m not sure why exactly, but I’d rather stick with existing process for now: https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/release/release-tasks.md#repeatedly-cut-a-release-1

furkatgofurov7: I wonder why, though, but it is fine. Thanks!

/lgtm

PR #206, opened 2023-08-04
🌱 Add a preflight check to validate core provider name

Fedosin:
When you try to install a Core Provider with a name different from "cluster-api", clusterctl validation [fails](https://github.com/kubernetes-sigs/cluster-api/blob/main/cmd/clusterctl/client/config/providers_client.go#L414-L416).
After that CAPI operator puts a vague error description in the logs and doesn't add anything meaningful to the provider status.

This PR adds a pre-flight check to compare the Core Provider name with the expected name, allowing for earlier failure and an informative update to the provider status.

Fixes #


alexander-demicev (approved):
/approve

damdo (commented):
/lgtm

PR #207, opened 2023-08-04
Refactor tests to be more readable and have better logging

Danil-Grigorev:
Adds a test framework package for helper methods

Fixes #


alexander-demicev (approved):
/approve

test/e2e/resources/full-chart-install.yaml
```diff
@@ -0,0 +1,6775 @@
+---
+# Source: cluster-api-operator/templates/operator-components.yaml
+apiVersion: apiextensions.k8s.io/v1
```
alexander-demicev, line 3: can this file be updated as a part of `make generate`? I assume the tests will break each time we add a new API change. It can be done in a follow-up PR

Fedosin (approved):
/lgtm

test/e2e/resources/full-chart-install.yaml
Danil-Grigorev: I fixed this in https://github.com/kubernetes-sigs/cluster-api-operator/pull/212

PR #211, opened 2023-08-10
🐛 Fix incorrect fieldSpecs paths

Fedosin:

Several paths for name references are not correct and therefore kustomize doesn't substitute right values there.

This PR fixes it and makes kustomize works correctly.

Reference patch: https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/config/crd/patches/webhook_in_coreproviders.yaml#L18

Fixes #

### DIFF
--- a/test/e2e/resources/full-chart-install.yaml
+++ b/test/e2e/resources/full-chart-install.yaml
@@ -16,7 +16,7 @@ spec:
       clientConfig:
         caBundle: Cg==
         service:
-          name: webhook-service
+          name: capi-operator-webhook-service
           namespace: 'capi-operator-system'
           path: /convert
       conversionReviewVersions:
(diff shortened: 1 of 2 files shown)


furkatgofurov7 (approved):
Merging it since it is straightforward change, thanks @Fedosin. 

/approve
/lgtm

Fedosin:
New values are correct, but it breaks the operator:
`Error from server (InternalError): error when creating "core.yaml": Internal error occurred: failed calling webhook "vcoreprovider.kb.io": failed to call webhook: the server could not find the requested resource`.

🤦

test/e2e/resources/full-chart-install.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       clientConfig:
         caBundle: Cg==
         service:
-          name: webhook-service
+          name: capi-operator-webhook-service
```
furkatgofurov7, line 19: where is this service being defined?

Fedosin: It is here https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/test/e2e/resources/full-chart-install.yaml#L6421

furkatgofurov7: /lgtm

PR #216, opened 2023-08-22
🐛 Fix imagePullSecrets bug

adriananeci:

Based on https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers `imagePullSecrets` is part of the PodSpec, not ContainerSpec

This PR fix this bug by moving the `imagePullSecrets` to the right indentation, and also adjusts the values files accordingly.

Without this fix, the helm chart is not usable 😞 

Fixes #

Maybe a tool like https://github.com/yannh/kubeconform ran as a CI step might help in catching similar issues in the future


JonnieDoe (commented):
/lgtm

PR #250, opened 2023-09-12
✨ Add provider healthcheck controller

Fedosin:
Currently we don't check the provider health after its manifests are applied, and Provider controller sets `Installed` and `Ready` status conditions to `True` at the same time. It may be confusing for users, especially when provider startup fails - User sees the provider CR with condition Ready=True, but the deployment is not available.

To fix this we introduce Provider Health Check Controller, which will check the deployment and update Ready status condition accordingly.

Fixes #

### DIFF
--- a/internal/controller/healthcheck/healthcheck_controller.go
+++ b/internal/controller/healthcheck/healthcheck_controller.go
@@ -1,0 +1,198 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package healthcheck
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	operatorv1 "sigs.k8s.io/cluster-api-operator/api/v1alpha2"
+	"sigs.k8s.io/cluster-api-operator/internal/controller/genericprovider"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util/conditions"
+	"sigs.k8s.io/cluster-api/util/patch"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type ProviderHealthCheckReconciler struct {
+	Client client.Client
+}
+
+const (
+	providerLabelKey = "cluster.x-k8s.io/provider"
+)
+
+func (r *ProviderHealthCheckReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&appsv1.Deployment{}, builder.WithPredicates(providerDeploymentPredicates())).
+		WithOptions(options).
+		Complete(r)
+}
+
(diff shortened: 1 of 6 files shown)


Fedosin:
/hold
for community discussion

Fedosin (commented):
/hold cancel

alexander-demicev (commented):
looks good to me, just one minor comment

internal/controller/healthcheck/healthcheck_controller.go
```diff
@@ -0,0 +1,203 @@
...
+			Status: corev1.ConditionFalse,
+		})
+	}
+
+	// Don't requeue immediately if the deployment is not ready, but rather wait 5 seconds.
+	if conditions.IsFalse(typedProvider, clusterv1.ReadyCondition) {
+		result = ctrl.Result{RequeueAfter: 5 * time.Second}
+	}
+
+	options := patch.WithOwnedConditions{Conditions: []clusterv1.ConditionType{clusterv1.ReadyCondition}}
```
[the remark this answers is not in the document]
Fedosin: Yeah, but since the spec stays the same, it stops almost immediately https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/internal/controller/genericprovider_controller.go#L121-L125

internal/controller/healthcheck/healthcheck_controller.go
```diff
@@ -0,0 +1,203 @@
...
+
+	result := ctrl.Result{}
+
+	deployment := &appsv1.Deployment{}
+
+	if err := r.Client.Get(ctx, req.NamespacedName, deployment); err != nil {
+		if apierrors.IsNotFound(err) {
+			// Object not found, return. Created objects are automatically garbage collected.
+			// For additional cleanup logic use finalizers.
+			return result, nil
```
furkatgofurov7, line 71: should not we return error here since object is anyways missing?

internal/controller/healthcheck/healthcheck_controller.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+		// Error reading the object - requeue the request.
+		return result, err
```
furkatgofurov7, line 74: I think we need to flip this with the above return statement?

internal/controller/healthcheck/healthcheck_controller.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	// Stop earlier if this provider is not fully installed yet.
+	if !conditions.IsTrue(typedProvider, operatorv1.ProviderInstalledCondition) {
+		return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
+	}
+
+	// Compare provider's Ready condition with the deployment's Available condition and stop if they already match.
+	currentReadyCondition := conditions.Get(typedProvider, clusterv1.ReadyCondition)
+	if currentReadyCondition != nil && deploymentCondition != nil && currentReadyCondition.Status == deploymentCondition.Status {
```
furkatgofurov7, line 98: Q: are not we comparing `True` with `Available` here?

internal/controller/healthcheck/healthcheck_controller.go
```diff
@@ -0,0 +1,203 @@
...
+		conditions.Set(typedProvider, &clusterv1.Condition{
+			Type:   clusterv1.ReadyCondition,
+			Status: corev1.ConditionFalse,
+		})
+	}
+
+	// Don't requeue immediately if the deployment is not ready, but rather wait 5 seconds.
+	if conditions.IsFalse(typedProvider, clusterv1.ReadyCondition) {
+		result = ctrl.Result{RequeueAfter: 5 * time.Second}
+	}
```
furkatgofurov7, line 124: Q: The sequence of var definition and there usage seems to be out of the way here. I think this whole block has to be moved up to line 85 or at least before initializing a patch helper?

furkatgofurov7 (commented):
@Fedosin thanks for PR!

Overall LGTM, few clarification questions on reconciliation handling logic

internal/controller/healthcheck/healthcheck_controller.go
Fedosin: I rewrote it as
```
	if err := r.Client.Get(ctx, req.NamespacedName, deployment); err != nil {
		// Error reading the object - requeue the request.
		return result, err
	}
```
Now we always return the error.

internal/controller/healthcheck/healthcheck_controller.go
Fedosin: No, we compare condition statuses, they are both "True" or "False". "Available" is a type:
```
  - lastTransitionTime: "2023-10-27T15:23:43Z"
    lastUpdateTime: "2023-10-27T15:23:43Z"
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
```

Fedosin: Also, to prevent confusion I renamed deploymentCondition -> deploymentAvailableCondition

internal/controller/healthcheck/healthcheck_controller.go
Fedosin: Maybe I'm wrong, but we can't do this before initializing the patch helper. When we create it, it stores the original value of the object. Then we do some modifications with the object. Finally, when we call Patch, the helper calculates a diff between the original value and the modified one.
If we move these modifications before initializing the helper, the diff will always be empty.

alexander-demicev (approved):
/approve

PR #256, opened 2023-09-21
:bug: Fix url fetch from gitlab repository

CharlieR-o-o-t:
Fixes bug with gitlab fetching.

url.RawPath is optional, and fulfilled only if provider 'fetchUrl' contains urlencoded characters.  It's better to use Url.Path


CharlieR-o-o-t:
Oh, in vendored cluster-api package is the same approach. 
Going to fix at the cluster-api side.

Fedosin (approved):
/approve

alexander-demicev (approved):
/lgtm

PR #335, opened 2023-11-23
🌱 Disable dependabot for hack/* modules

Fedosin:
It's not necessary to update dependency for internal tools so frequently. So, it's better to disable dependabot for them, or otherwise it creates to many PRs.

Fixes #


furkatgofurov7:
>It's not necessary to update dependency for internal tools so frequently. So, it's better to disable dependabot for them, or otherwise it creates to many PRs.


If it is just too frequent but we still need updates, we could increase the interval of updates done by the bot and keep it running, otherwise, let's remove it completely

alexander-demicev:
can we do monthly updates? For test package also maybe

furkatgofurov7 (approved):

alexander-demicev (approved):
/lgtm

PR #341, opened 2023-12-01
🌱  Move generic provider interface to API package

Danil-Grigorev:

This PR moves generic provider interface into API package. This allows to drop wrapper objects from internal package.

Fixes #


api/v1alpha2/infrastructureprovider_wrapper.go
alexander-demicev, line 1: should we name these files `*provider_wrapper.go?

controllers/alias.go
alexander-demicev, line 1: I'm ok with dropping this alias, not sure if anyone is using a controller like this

alexander-demicev (approved):

Fedosin:
Hmm... I tried to do something similar, but @JoelSpeed wanted to keep api package as simple as possible.  I'll let him leave a comment.

Danil-Grigorev:
This change does not bring additional API dependencies from the currently used set.

Fedosin (approved):
/lgtm

PR #349, opened 2023-12-06
✨Allow to utilize providers from clusterctl.yaml file in /config/clusterctl.yaml

Danil-Grigorev:

This allows to pass a default clusterctl.yaml file to the operator to fetch provider list from.
The file should be located under `/config/clusterctl.yaml`.

Fixes #

### DIFF
--- a/internal/controller/consts.go
+++ b/internal/controller/consts.go
@@ -27,4 +27,5 @@ const (
 	githubDomain            = "github.com"
 	gitlabHostPrefix        = "gitlab."
 	gitlabPackagesAPIPrefix = "/api/v4/projects/"
+	configPath              = "/config/clusterctl.yaml"
 )
--- a/internal/controller/phases.go
+++ b/internal/controller/phases.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"os"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


Fedosin (commented):
Default providers are hardcoded in clusterctl: https://github.com/kubernetes-sigs/cluster-api/blob/9d36ddcbf8381574a86cebeacdcdfc85bafa71e9/cmd/clusterctl/client/config/providers_client.go#L128

For a custom provider you should append its URL to the FetchConfig inside Provider CR spec. https://github.com/kubernetes-sigs/cluster-api-operator/tree/main/docs#provider-spec

Can you elaborate more on how you want to use clusterctl.yaml file?

Danil-Grigorev:
`Clusterctl.yaml` has a [list of repositories](https://cluster-api.sigs.k8s.io/clusterctl/configuration.html?highlight=clusterctl%20config#provider-repositories), which extend the default list and can provide a CAPI Operator installation a list of “non-default” repositories to use, without a need to specify fetch-config for everything which is not in the hardcoded list of defaults bound to a release version. The file can be mounted as a ConfigMap, and this supports autoreload, so this list could be constantly up-to-date, disregarding the release. Additionally it allows to bound a default for a set of provider manifests to specific version, instead of `latest`.

Fedosin:
I would recommend against it. It complicates the logic of the operator and makes it harder to troubleshoot potential issues. It seems to me that the current way when you can declare custom providers only in the CR specification is the best way.

@alexander-demicev @furkatgofurov7 @damdo your opinions?

alexander-demicev (approved):
/approve

Fedosin (approved):
/lgtm

PR #352, opened 2023-12-11
🐛 Prevent propagation of OwnerReferences from provider manifests

### LINKED ISSUE #350: CoreProvider never becomes ready
[A clear and concise description on how to REPRODUCE the bug.]

- Install capi-operator v0.7.0
- Apply a `CoreProvider` manifest.
- Apply any other provider manifest.
No provider is getting provisioned, as they are failing pre-flight check, expecting
```
    status: "True"
    type: Ready
```
condition on the `CoreProvider` instance.

Core provider becoming ready and allowing the rest of the providers to start provisioning.

**Environment:**

- Cluster-api-operator version: v0.7.0
- Cluster-api version: any
- Minikube/KIND version: any
- Kubernetes version: (use `kubectl version`): any
- OS (e.g. from `/etc/os-release`): any

/kind bug

Danil-Grigorev:

Healthcheck implementation expects deployment availability and an ownership reference pointing to the owner resource, to mark the provider as ready. However if the provider manifest is owned by some other resource, the second ownership reference appears on the deployment too, which is unexpected.

Fixes #350

### DIFF
--- a/internal/controller/component_customizer.go
+++ b/internal/controller/component_customizer.go
@@ -58,7 +58,12 @@ func customizeObjectsFn(provider genericprovider.GenericProvider) func(objs []un
 
 			if o.GetNamespace() != "" {
 				// only set the ownership on namespaced objects.
-				o.SetOwnerReferences(util.EnsureOwnerRef(provider.GetOwnerReferences(),
+				ownerReferences := o.GetOwnerReferences()
+				if ownerReferences == nil {
+					ownerReferences = []metav1.OwnerReference{}
+				}
+
+				o.SetOwnerReferences(util.EnsureOwnerRef(ownerReferences,
 					metav1.OwnerReference{
 						APIVersion: operatorv1.GroupVersion.String(),
 						Kind:       provider.GetObjectKind().GroupVersionKind().Kind,


Fedosin:
Hey! Probably we need to fix this issue here: https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/internal/controller/healthcheck/healthcheck_controller.go#L71-L76

Probably it's okay to have multiple owners for deployments, but there should be only one Provider owner. So, if we have just one owner, we need to add a check that this owner is a Prover CR. Otherwise we have to iterate over the owners and find the Provider one (or return an error if nothing found)

Danil-Grigorev:
> Hey! Probably we need to fix this issue here: https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/internal/controller/healthcheck/healthcheck_controller.go#L71-L76
> 
> Probably it's okay to have multiple owners for deployments, but there should be only one Provider owner. So, if we have just one owner, we need to add a check that this owner is a Prover CR. Otherwise we have to iterate over the owners and find the Provider one (or return an error if nothing found)

I tried to refactor the reconciler which will fix this issue. Here is a separate PR for that: https://github.com/kubernetes-sigs/cluster-api-operator/pull/354

Fedosin (approved):
/approve

Fedosin:
The patch is good, thanks for fixing the bug!

PR #356, opened 2023-12-14
:book: README: Added AddonProvider.

guettli:

`AddonProvider` was missing at some places.


alexander-demicev (approved):
/approve
/lgtm

PR #370, opened 2023-12-26
✨Add IPAM provider support

### LINKED ISSUE #367: Support of cluster api IPAM provider
**User Story**

As an operator I would like cluster api IPAM provider to be supported for beeing able to deploy all providers I need using cluster api operator only (and not use clusterctl for IPAM provider and cluster api operator for other providers).

**Detailed Description**

Cluster api IPAM provider is a specific provider supported by cluster api but currently not supported in cluster api operator for beeing deployed. It will be great if you can add support for this provider too.

/kind feature

Fedosin:

This PR adds a new provider type - `IPAMProvider`, that will be used to manage Cluster API [IPAM providers](https://cluster-api.sigs.k8s.io/reference/providers.html#ip-address-management-ipam).

Fixes #367


Danil-Grigorev (commented):
/lgtm

api/v1alpha2/ipamprovider_types.go
```diff
@@ -0,0 +1,59 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
furkatgofurov7, line 2: nit: header year has to be updated here in other places

furkatgofurov7 (approved):

api/v1alpha2/ipamprovider_types.go
Fedosin: Idk... the code was written last year and hasn't been updated after, so I think it is correct to put 2023 there.

PR #395, opened 2024-01-17
⚠️ Add short names for custom resources

### LINKED ISSUE #361: Support short Names for capi-operator objects
**User Story**

As an operator, I want to use shortNames while managing cluster-api-operator objects in the cluster. As of now, we don't support any shortnames for any of our custom resources. Given the names are large, it will be great if we add short names. 

```bash
NAME                      SHORTNAMES   APIVERSION                           NAMESPACED   KIND
addonproviders                         operator.cluster.x-k8s.io/v1alpha2   true         AddonProvider
bootstrapproviders                     operator.cluster.x-k8s.io/v1alpha2   true         BootstrapProvider
controlplaneproviders                  operator.cluster.x-k8s.io/v1alpha2   true         ControlPlaneProvider
coreproviders                          operator.cluster.x-k8s.io/v1alpha2   true         CoreProvider
infrastructureproviders                operator.cluster.x-k8s.io/v1alpha2   true         InfrastructureProvider
```

**Detailed Description**

I think we can use `p` for provider given it's common in all along with type of provider in the prefix. For example, something like `addonp` for `addonprovider` and `infrap` for `infraprovider` but even after reduction this sounds a little big. :)
I'm open to more thoughts here. 

Happy to work on this once we finalize on the naming convention. 

/kind feature

Fedosin:

Fixes #361

### DIFF
--- a/api/v1alpha2/addonprovider_types.go
+++ b/api/v1alpha2/addonprovider_types.go
@@ -31,6 +31,7 @@ type AddonProviderStatus struct {
 }
 
 // +kubebuilder:object:root=true
+// +kubebuilder:resource:path=addonproviders,shortName=caap,scope=Namespaced
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="InstalledVersion",type="string",JSONPath=".status.installedVersion"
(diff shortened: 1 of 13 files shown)


alexander-demicev (approved):
/approve

kranurag7 (approved):

PR #400, opened 2024-01-18
⚠️ Deprecate v1alpha1

Fedosin:

Fixes #


Danil-Grigorev (commented):
I think we need to include changes for v1alpha2 in helm charts as well. So `secretName`/`secretNamespace` wouldn't appear that much anymore. We store v1alpha2 only now, regardless of the version.

alexander-demicev (approved):
/lgtm

PR #407, opened 2024-01-25
✨ Add a preflight check to prevent provider downgrades

Fedosin:

Cluster API doesn't support provider downgrades due to potential breaking changes in CRDs. Currently, if users try to downgrade for example from v1.6.1 to v1.5.2, they get a error:

```
invalid provider metadata: version v1.6.1 (the current version) for the provider capi-system/cluster-api does not match any release series
```
This is not really informative, so it is better to add a preflight check to fail earlier and provide more descriptive error messages.

Fixes #


Danil-Grigorev (approved):
/lgtm

alexander-demicev (approved):
/approve

PR #422, opened 2024-02-14
🐛 Restrict configmap lookup to the provider namespace

Fedosin:

When we use configmaps as a provider source, we filter them by provided labels and then take the latest version. Unfortunately, we list configmaps in all namespaces, which is not correct.

This PR restricts the search by the provider namespace only.

Fixes #


Danil-Grigorev (commented):
/lgtm
Thank you

alexander-demicev (approved):
/approve

PR #453, opened 2024-03-07
don’t deliver cert-manager CRDs with cluster-api-operator chart

### LINKED ISSUE #281: Installing the chart creates cert-manager CRDs
`helm install --repo https://kubernetes-sigs.github.io/cluster-api-operator capi-operator cluster-api-operator -n capi-system --create-namespace`

That the operator would get installed, nothing more.

I understand that the operator might depend on cert-manager, but then it should just depend on it (which is currently not possible because of https://github.com/cert-manager/cert-manager/issues/6179#issuecomment-1764051690) and otherwise just shouldn't do anything with the cert-manager deployment and just require the user to install it manually. Because now I can't install cert-manager my way, as the CRDs already exist 😅, I'd have to `skipCRDs` in my gitops system.

I guess this isn't immediately fixable, but then this issue is just to track the problem 😁

**Environment:**

- Cluster-api-operator version: N/A
- Cluster-api version: N/A
- Minikube/KIND version: N/A
- Kubernetes version: (use `kubectl version`): N/A
- OS (e.g. from `/etc/os-release`): N/A
- Helm Chart version: 0.6.0 <- missing from the template

/kind bug
/area helm-chart <- I couldn't find any fitting area

jackfrancis:

This PR changes the way that the cluster-api-operator chart relates to the cert-manager chart. Rather than package those downstream CRDs with the chart, we can simply dispatch in the cluster-api-operator chart the `cert-manager.installCRDs=true` value config.

So, with this PR, a simple `helm install --repo https://kubernetes-sigs.github.io/cluster-api-operator capi-operator cluster-api-operator` command will no longer include the cert-manager CRDs as part of its resource payload.

To explicitly install the cert-manager chart, including the CRDs that are delivered with the cert-manager chart, as part of the cluster-api-operator helm install, we will do:

- `helm install --repo https://kubernetes-sigs.github.io/cluster-api-operator capi-operator cluster-api-operator --set cert-manager.enabled=true`

If a user wishes to install the cert-manager chart **without the CRDs**, this is the command:

- `helm install --repo https://kubernetes-sigs.github.io/cluster-api-operator capi-operator cluster-api-operator --set cert-manager.enabled=true --set cert-manager.installCRDs=false`

Fixes #281

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -476,7 +476,7 @@ release-manifests: $(KUSTOMIZE) $(RELEASE_DIR) ## Builds the manifests to publis
 release-chart: $(HELM) $(KUSTOMIZE) $(RELEASE_DIR) $(CHART_DIR) $(CHART_PACKAGE_DIR) ## Builds the chart to publish with a release
 	cp -rf $(ROOT)/hack/charts/cluster-api-operator/. $(CHART_DIR)
 	$(KUSTOMIZE) build ./config/chart > $(CHART_DIR)/templates/operator-components.yaml
-	$(ROOT)/hack/inject-cert-manager-helm.sh $(CERT_MANAGER_VERSION)
+	$(ROOT)/hack/inject-cert-manager-chart-version.sh $(CERT_MANAGER_VERSION)
 	$(HELM) package $(CHART_DIR) --app-version=$(HELM_CHART_TAG) --version=$(HELM_CHART_TAG) --destination=$(CHART_PACKAGE_DIR)
 
 .PHONY: release-staging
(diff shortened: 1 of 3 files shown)


alexander-demicev (approved):
/approve

Danil-Grigorev:
/lgtm
Thanks, makes sense to me.

zioproto:
Is this change included in chart version 0.9.2 ? Or it needs to be cherry-picked ?

zioproto:
@jackfrancis could we cherry-pick this change to have it available in the helm chart ? thanks

furkatgofurov7:
> Is this change included in chart version 0.9.2 ? Or it needs to be cherry-picked ?

@zioproto hey, v0.9.2 is already released and this change was not backported and so not part of that patch release. I think it is fine to be backported and include it in the coming v0.9.3 patch release

/cherrypick release-0.9

PR #467, opened 2024-03-21
🌱 Use 1.21 golang version for tools

Danil-Grigorev:
Test-infra jobs are already using 1.21 under the hood, so we can’t use go 1.20 for jobs here. This causes `dependabot_pr.yaml` to create a conflicting change in `make verify` as opposed to current main.

Fixes #


alexander-demicev (approved):
/approve
/lgtm

PR #475, opened 2024-03-25
[WIP] (rebasing) ✨ Introduce generics to providers

Danil-Grigorev:

This implementation achieves the following:
- Separate each provider in it’s own reconciler.
- Share common logic between reconcilers via phase functions, while preserving the original provider type per reconciler. Reconcilers can reuse common methods without restrictions or add their own, applicable only for the reconciler itself.
- Unifies Clusterctl provider generation for plugin and provider reconcilers under `providers` package import. Removes the need to do switch cases on provider types in multiple locations. Only one needs to be up-to-date now.
- Reconcile is split into stages. These are defined like this:
```golang
	PreflightChecks(ctx context.Context, provider P) []ReconcileFn[P, Group[P]]
	ReconcileNormal(ctx context.Context, provider P) []ReconcileFn[P, Group[P]]
	ReportStatus(ctx context.Context, provider P) []ReconcileFn[P, Group[P]]
	ReconcileDelete(ctx context.Context, provider P) []ReconcileFn[P, Group[P]]

```
Each provider can extend/replace/re-order the phases for each stage at their liking.

Additionally a couple of issues are fixed:
- Core provider singleton check. Previously Core provider was using name check for differentiating other core providers, but the core provider can only be named `cluster-api`.
- Pre-flight checks are now unique per each provider, type casting to check provider type to perform some validation is no longer needed.

Fixes #


furkatgofurov7:
@Danil-Grigorev can you please rebase this PR if this change is needed and to get things going?

anmazzotti (approved):

Danil-Grigorev:
@furkatgofurov7 can you please help with reviewing this change? I'm planning to rebase it soon top of recent changes.

This should be the entry point to improve operator memory and performance footprint

furkatgofurov7:
> @furkatgofurov7 can you please help with reviewing this change? I'm planning to rebase it soon top of recent changes.
> 
> This should be the entry point to improve operator memory and performance footprint

@Danil-Grigorev please rebase, overall the direction of the changes look good to me,  it might slightly deviate after the rebase so I will take another look after

Danil-Grigorev:
I so far pushed a version which sits on top of a current `main`, but it is nowhere near being working. I will try to spend time on making it operational again later. Will mark it as a WIP for now.

PR #500, opened 2024-04-24
Add helm.sh/resource-policy: keep to CRDs to prevent deletion

Danil-Grigorev:

When CAPI Operator chart used as a dependency, the removal of the chart hosting the dependency causes CRDs to be removed. This happens because CRDs are generated as a part of `base` and therefore are managed by helm 3. Helm [docs](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource) define `keep` policy as:
> skip deleting this resource when a helm operation (such as helm uninstall, helm upgrade or helm rollback) **would result in its deletion**

Setting this annotation should not prevent CRD upgrades, when resource is changed only.

Typically this removal happens after the main deployment is removed, and before any or all of the `*Provider` resources are removed. Provider resources has a finalizer, nothing can remove at this point, so the deletion hangs and later fails. Happens because CRD removal requires all resources it defines to be cleaned up before being removed itself.

Fixes #

### DIFF
--- a/config/chart/kustomization.yaml
+++ b/config/chart/kustomization.yaml
@@ -23,6 +23,11 @@ bases:
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
 #- ../prometheus
 
+patches:
+- path: patches/keep-crds.yaml
+  target:
+    kind: CustomResourceDefinition
+
 patchesStrategicMerge:
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
 # Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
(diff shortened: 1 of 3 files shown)


Fedosin (commented):
For chart we have a dedicated section: https://github.com/kubernetes-sigs/cluster-api-operator/tree/main/config/chart
Maybe it's better to put your patch there, since it affects the chart only?

Danil-Grigorev:
I added it there @Fedosin, although in this case an additional annotation makes difference for helm only anyway. Helm does not care what is on resource, it needs this annotation in the chart resources, so it can’t do anything regardless of the placement

Danil-Grigorev:
I performed some testing of the helm upgrade, everything works with this change, so the docs are not wrong. Unless the resource is removed, this annotation changes nothing.

alexander-demicev (approved):
/approve

furkatgofurov7 (commented):
/lgtm
/hold

To avoid merging, in case @Fedosin has some more questions, otherwise unhold

Fedosin (approved):
/lgtm
/hold cancel

PR #515, opened 2024-05-02
🐛 Revert "don’t deliver cert-manager CRDs with cluster-api-operator chart"

Fedosin:
This reverts commit c8a9022ba90e00c259ef5a8c4ca6a69d69180ed4.

PR #453 broke installation of cert-manager CRDs, so we revert that regression.

The reason why cert-manager CRDs were placed in "crds" folder, is because in this case Helm [applies](https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds) them earlier, before all other manifests. Also Helm guarantees that if these CRDs are already presented in the system, it won't [overwrite](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you) them.

If we template everything together, both cert-manager resources and CRDs will be applied simultaneously, and we get this error:
```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "capi-operator-serving-cert" namespace: "capi-operator-system" from "": no matches for kind "Certificate" in version "cert-manager.io/v1" ensure CRDs are installed first, resource mapping not found for name: "capi-operator-selfsigned-issuer" namespace: "capi-operator-system" from "": no matches for kind "Issuer" in version "cert-manager.io/v1" ensure CRDs are installed first]
``` 

Fixes #


Fedosin:
During the community meeting we decided to remove cert-manager support from the helm chart, because it is doesn't work with ArgoCD.

PR #519, opened 2024-05-07
⚠️ feat(rbac): use real ServiceAccount instead of default

### LINKED ISSUE #518: Use real ServiceAccount instead of `default`
I tried installing the chart, failing at our policy to not use the default ServiceAccount

That the chart would create it's own ServiceAccount for the operator.

It is best-practice to create specific ServiceAccounts for each component.

**Environment:**

- Cluster-api-operator version: latest
- Cluster-api version: N/A
- Minikube/KIND version: N/A
- Kubernetes version: (use `kubectl version`): N/A
- OS (e.g. from `/etc/os-release`): N/A

/kind bug

cwrau:
Closes #518

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -15,6 +15,8 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      serviceAccountName: manager
+      automountServiceAccountToken: true
       containers:
       - command:
         - /manager
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -3,3 +3,4 @@ resources:
 - role_binding.yaml
 - leader_election_role.yaml
 - leader_election_role_binding.yaml
+- service_account.yaml
--- a/config/rbac/leader_election_role_binding.yaml
+++ b/config/rbac/leader_election_role_binding.yaml
@@ -8,5 +8,5 @@ roleRef:
   name: leader-election-role
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: manager
   namespace: system
--- a/config/rbac/role_binding.yaml
+++ b/config/rbac/role_binding.yaml
@@ -8,5 +8,5 @@ roleRef:
   name: manager-role
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: manager
   namespace: system
--- a/config/rbac/service_account.yaml
+++ b/config/rbac/service_account.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: manager
+  namespace: system
--- a/hack/charts/cluster-api-operator/templates/deployment.yaml
+++ b/hack/charts/cluster-api-operator/templates/deployment.yaml
@@ -47,6 +47,8 @@ spec:
       {{- toYaml . | nindent 8 }}
       {{- end }}
     spec:
+      serviceAccountName: capi-operator-manager
+      automountServiceAccountToken: true
       {{- with .Values.securityContext }}
       securityContext:
       {{- toYaml . | nindent 8 }}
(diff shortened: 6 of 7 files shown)


cwrau:
I don't know how I would make the `serviceAccountName` in the `deployment.yaml` "generated" like the name in the `service_account.yaml`, so I just hardcoded it to the value that got generated?

Fedosin (commented):
/ok-to-test

cwrau:
Heyho, is there something I can do to help this get merged? 😊

alexander-demicev (approved):
/approve

config/manager/manager.yaml
```diff
@@ -15,6 +15,8 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
+      serviceAccountName: manager
+      automountServiceAccountToken: true
```
furkatgofurov7, line 19: Q: is this set to false and best practice, when the default SA is used?

cwrau: If I understand the question correctly, yes, it's best-practice to set this to false when not using a custom serviceAccount to prevent the accidental usage of the default serviceAccount

In general it's best-practice to not use the default serviceAccount anyways

furkatgofurov7:
@cwrau since this is breaking change and can't be released with patch (and in fact it was not backported). I just checked the [changelog](https://github.com/kubernetes-sigs/cluster-api-operator/compare/v0.11.0...main) since last minor v0.11 release and it is quite small for cutting a minor release. How urgent this change for you, can we give some more time for minor release before cutting? 

cc @Fedosin @alexander-demicev @Danil-Grigorev

cwrau:
> @cwrau since this is breaking change 

Why is this a breaking change? 🤔
Everything just worked after this change for me

> and can't be released with patch release (and in fact it was not backported). I just checked the [changelog](https://github.com/kubernetes-sigs/cluster-api-operator/compare/v0.11.0...main) since last minor v0.11 release 

But if we assume that this is a breaking change, why would you ever want to release this as a minor version instead of a breaking one?! 😮

> and it is quite small for cutting a minor release. How urgent this change for you, can we give some more time for new minor release before cutting it?
> 
> cc @Fedosin @alexander-demicev @Danil-Grigorev

Is it a problem to release frequently with small changes? 🤔

But it's not urgent right now, we're using flux pathes to rectify this in the meantime

furkatgofurov7:
> Why is this a breaking change? 🤔
> Everything just worked after this change for me

That is because the PR was labelled with ⚠️: https://github.com/kubernetes-sigs/cluster-api-operator/blob/fbe3b5a212577abece3350d2557d8496a9721de4/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L2 so that was my assumption 

>But if we assume that this is a breaking change, why would you ever want to release this as a minor version instead of a breaking one?!
Breaking changes usually accepted to be releases as part of the minor releases and not the patch. I am not sure what do you mean as "breaking one" in this context.

cwrau:
> > Why is this a breaking change? 🤔
> > Everything just worked after this change for me
> 
> That is because the PR was labelled with ⚠️:
> 
> https://github.com/kubernetes-sigs/cluster-api-operator/blob/fbe3b5a212577abece3350d2557d8496a9721de4/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L2
> 
> so that was my assumption

Ah, @Fedosin changed the title at some point, maybe he can chime in why?

> > But if we assume that this is a breaking change, why would you ever want to release this as a minor version instead of a breaking one?!
> 
> Breaking changes are usually accepted and released as part of the minor releases and not the patch. I am not sure what do you mean as "breaking one" in this context.

Oh, sorry, typo 😅
I meant major release

So this project is not following semver? 😕

furkatgofurov7:
> So this project is not following semver? 😕

Well, it does follow semver in a way similar to how CAPI does https://cluster-api.sigs.k8s.io/contributing?highlight=release#breaking-changes, but in semVer world breaking changes should land in major release, that I agree. 

Also, by looking at this PR again, IMO it is not a breaking change but rather a bug fix considering what it tries to solve

PR #521, opened 2024-05-08
⚠️ remove cert-manager support from helm chart

Fedosin:
Right now we have experimental support for Cert Manager installation in the chart. As part of this, we store Cert Manager CRDs in the "crds" folder. If you install the chart with Helm, everything works fine, resources are only installed if they don't exist in the cluster. But if we use Argo, we always install CRDs even if they already exist by overwriting them. Since we can't get around this limitation, we decided to remove Cert Manager support from the chart.

We recommend using the clusterctl [plugin](https://cluster-api-operator.sigs.k8s.io/02_installation/01_plugin-installation) for local development and testing that retains support for installing Cert Manager. In the case of the chart, users must manage Cert Manager installation themselves.

Fixes #


Fedosin:
/hold
for further duscussions.

alexander-demicev (approved):
/approve

PR #536, opened 2024-05-24
:sparkles: Add featureGates to ControlPlaneProvider

kahirokunn:

This PR adds support for specifying featureGates in the ControlPlaneProvider. This is particularly necessary when using infrastructure=docker, as we need to pass CLUSTER_TOPOLOGY=true to the kubeadm provider. Currently, there is no way to do this, and this feature is essential for enabling cluster topology features in kubeadm.

**Usage Example**:

To use this feature, you can install the capi-operator with Helm and specify the feature gates in your values file. Here is an example:

```bash
helm install capi-operator cluster-api-operator/cluster-api-operator --version 0.10.1 -n capi-operator-system --create-namespace --values ./example-values.yaml
```

`example-values.yaml`:
```yaml
cert-manager:
  enabled: false
  installCRDs: false
infrastructure: docker:v1.7.2
core: cluster-api:v1.7.2
controlPlane: kubeadm:v1.7.2
manager:
  featureGates:
    core:
      ClusterTopology: true
      MachinePool: true
    docker:
      ClusterTopology: true
      MachinePool: true
    kubeadm:
      ClusterTopology: true
      MachinePool: true
```

**Sample Output**:

Here is a portion of the expanded manifest showing the applied feature gates:

```yaml
# Source: cluster-api-operator/templates/control-plane.yaml
apiVersion: operator.cluster.x-k8s.io/v1alpha2
kind: ControlPlaneProvider
metadata:
  name: kubeadm
  namespace: kubeadm-control-plane-system
  annotations:
    "helm.sh/hook": "post-install"
    "helm.sh/hook-weight": "2"
spec:
  version: v1.7.2
  manager:
    featureGates:
      ClusterTopology: true
      MachinePool: true
```

Fixes #

### DIFF
--- a/hack/charts/cluster-api-operator/templates/infra-conditions.yaml
+++ b/hack/charts/cluster-api-operator/templates/infra-conditions.yaml
@@ -53,6 +53,15 @@ metadata:
     "argocd.argoproj.io/sync-wave": "2"
 {{- with .Values.configSecret }}
 spec:
+{{- if $.Values.manager }}
+  manager:
+{{- if and $.Values.manager.featureGates $.Values.manager.featureGates.kubeadm }}
+    featureGates:
+    {{- range $key, $value := $.Values.manager.featureGates.kubeadm }}
+      {{ $key }}: {{ $value }}
+    {{- end }}
+{{- end }}
+{{- end }}
   configSecret:
     name: {{ .name }}
     {{- if .namespace }}
--- a/hack/charts/cluster-api-operator/templates/control-plane.yaml
+++ b/hack/charts/cluster-api-operator/templates/control-plane.yaml
@@ -38,11 +38,24 @@ metadata:
   annotations:
     "helm.sh/hook": "post-install,post-upgrade"
     "helm.sh/hook-weight": "2"
-{{- if or $controlPlaneVersion $.Values.configSecret.name }}
+{{- if or $controlPlaneVersion $.Values.configSecret.name $.Values.manager }}
 spec:
 {{- end}}
 {{- if $controlPlaneVersion }}
   version: {{ $controlPlaneVersion }}
+{{- end }}
+{{- if $.Values.manager }}
+  manager:
+{{- if hasKey $.Values.manager.featureGates $controlPlaneName }}
+{{- range $key, $value := $.Values.manager.featureGates }}
+  {{- if eq $key $controlPlaneName }}
+    featureGates:
+    {{- range $k, $v := $value }}
+      {{ $k }}: {{ $v }}
+    {{- end }}
+  {{- end }}
+{{- end }}
+{{- end }}
 {{- end }}
 {{- if $.Values.configSecret.name }}
   configSecret:
--- a/test/e2e/helm_test.go
+++ b/test/e2e/helm_test.go
@@ -261,4 +261,20 @@ var _ = Describe("Create a proper set of manifests when using helm charts", func
 		Expect(err).ToNot(HaveOccurred())
 		Expect(manifests).To(MatchYAML(string(expectedManifests)))
 	})
+	It("should deploy kubeadm control plane with manager specified", func() {
+		manifests, err := helmChart.Run(map[string]string{
+			"core":           "cluster-api",
+			"controlPlane":   "kubeadm",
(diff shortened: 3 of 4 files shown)


hack/charts/cluster-api-operator/templates/infra-conditions.yaml
```diff
@@ -49,6 +49,15 @@ metadata:
     "helm.sh/hook-weight": "2"
 {{- with .Values.configSecret }}
 spec:
+{{- if $.Values.manager }}
+  manager:
+{{- if $.Values.manager.featureGates.kubeadm }}
```
alexander-demicev, line 54: can we make `kubeadm` configurable here? for users that are not using kubeadm for bootstrapping

alexander-demicev (commented):
thanks a lot for this PR, can I you to add tests here https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/test/e2e/helm_test.go?

hack/charts/cluster-api-operator/templates/infra-conditions.yaml
kahirokunn: @alexander-demicev 
Sorry, I don't seem to understand correctly what you mean.
Could you please explain in more detail?
Thx.

kahirokunn: As changes may be made, the test is written after hearing the explanation.
Thx.

kahirokunn: @alexander-demicev remind 🙇‍♂️

furkatgofurov7: @kahirokunn current implementation always assumes that `manager.featureGates` dictionary will use `kubeadm` as a bootstrap provider and nothing else. However, there is a possibilty that, other bootstrap provider, like `rke2` could be used instead of `kubeadm`, so making   `manager.featureGates` configurable would makes sense here only if we could make the whole `ControlPlaneProvider` type (starting from line 42) configurable including` metadata.name` & `metadata.namespace`. 

@alexander-demicev perhaps we could keep it as is, since it is already kubeadm specific, WDYT?

kahirokunn: I wrote this because the file is specific to kubeadm
 I wrote this because it is a kubeadm-specific file, and I thought it would be confusing if I wrote it in a standard way.

furkatgofurov7:
LGTM, but CI was not agreeing with me. Let's wait if retrigger helps (assuming it was a CI flake).

furkatgofurov7:
> LGTM, but CI was not agreeing with me. Let's wait if retrigger helps (assuming it was a CI flake).

We would need to check in template if feature gates are set before trying to access them, so that tests won't fail

furkatgofurov7:
@kahirokunn can you rebase on top of the main and look at the CI failure? This is almost ready to take in IMO

kahirokunn:
@furkatgofurov7 I fixed it by checking for the existence of the value so that the nil pointer exception would not occur.
I'd be happy if we could merge!

furkatgofurov7:
@kahirokunn this looks good already, the only pending comment that needs addressing is https://github.com/kubernetes-sigs/cluster-api-operator/pull/536#pullrequestreview-2095806454

kahirokunn:
@furkatgofurov7  
I have resolved the comment as the content has already been addressed.  
Would this be acceptable to you?

kahirokunn:
I understand.
I'll add it.

kahirokunn:
@furkatgofurov7 Thank you for pointing that out! I’ve written the necessary test and updated the file at test/e2e/helm_test.go as per the review comment. Please let me know if there’s anything else that needs to be addressed.

furkatgofurov7:
> @furkatgofurov7 Thank you for pointing that out! I’ve written the necessary test and updated the file at test/e2e/helm_test.go as per the review comment. Please let me know if there’s anything else that needs to be addressed.

Thanks for your patience and work on this PR, and sorry for late reply (I was on PTO), this looks good to me.

/approve

kahirokunn:
Thxs!

PR #538, opened 2024-05-31
✨ Add support for customizing additional provider deployments

willie-yao:
This PR adds support for customizing additional provider deployments via the `additionalDeployments` field in the provider spec. `additionalDeployments` is a map of the name of the provider to both the deployment and manager spec which is used for customizing the deployment/management container.

This PR also adds allows CRDPattern to be configurable in deployment customization. This is to allow for installing additional CRDs through the ASO deployment.

Fixes #


willie-yao:
I'm a bit confused on how to add a conversion function for CRDPattern. @Fedosin @alexander-demicev is there an example on how to do this? I tried adding something like this to the conversion funcs which didn't seem to work: https://github.com/willie-yao/cluster-api-operator/blob/upstream-install-aso-crds/api/v1alpha1/provider_conversion.go#L48

alexander-demicev:
@willie-yao You can find an example here https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4228/files, this PR added a new field to API and conversion to older versions

alexander-demicev:
@willie-yao This change is too specific to capz. What do you think about making it more generic? Can we introduce a new field to customize additional deployments that come with providers? I think something like `AdditionalDeployments` might work, where we can store the deployment names to customize and `DeploymentSpec` structure to apply these customizations. Do you think it makes sense?
I can see that in the future there might be more providers like CAPZ with more than one Deployment so it might be better to try to fix the problem in a bigger scope.

willie-yao:
@alexander-demicev That is a great idea. I did notice that this may be too specific to CAPZ to be implemented like this. I'll work on adding a field like `AdditionalDeployments`. Thanks for the suggestion!

api/v1alpha2/provider_types.go
```diff
@@ -73,6 +73,22 @@ type ProviderSpec struct {
...
+}
+
+type AdditionalProviderCustomization struct {
+	// Manager defines the properties that can be enabled on the controller manager for the provider.
+	// +optional
+	Manager *ManagerSpec `json:"manager,omitempty"`
+
+	// Deployment defines the properties that can be enabled on the deployment for the provider.
+	// +optional
+	Deployment *DeploymentSpec `json:"deployment,omitempty"`
```
willie-yao, line 91: @alexander-demicev not sure if this is the best way to do it, but I refactored the code to include an `AdditionalDeployments` field that maps the name of the deployment to the Manager and Deployment spec, where the additional deployments can be customized. Is this a good way to go about it?

hack/charts/cluster-api-operator/templates/infra.yaml
```diff
@@ -1,3 +1,13 @@
+{{- define "recursivePrinter" }}
+{{- range $key, $value := . }}
+{{- if kindIs "map" $value }}
+  {{ $key }}:
+  {{- include "recursivePrinter" $value | indent 2 }}
+{{- else }}
+  {{ $key }}: {{ $value }}
+{{- end }}
```
willie-yao, line 8: This function allows the chart to print out `additionalDeployments` correctly in yaml format as both deployments and manager has nested fields.

internal/controller/component_customizer.go
```diff
@@ -73,25 +73,41 @@ func customizeObjectsFn(provider operatorv1.GenericProvider) func(objs []unstruc
 					}))
 			}
 
+			//nolint:nestif
```
willie-yao, line 76: Not sure if a nolint is allowed here. I'll have to create a few helper functions if not. lmk which approach is better!

api/v1alpha2/provider_types.go
```diff
@@ -123,6 +141,11 @@ type ManagerSpec struct {
 	// in as container args to the provider's controller manager.
 	// Controller Manager flag is --feature-gates.
 	FeatureGates map[string]bool `json:"featureGates,omitempty"`
+
+	// CRDPattern is a regular expression that matches the CRDs that the provider
```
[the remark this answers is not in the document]
willie-yao: Yup, confirmed it works via container arguments in the deployment spec!

test/e2e/resources/only-infra-with-additional-deployments.yaml
```diff
@@ -0,0 +1,104 @@
...
+    "helm.sh/hook-weight": "2"
+    "argocd.argoproj.io/sync-wave": "2"
+spec:
+  configSecret:
+    name: test-secret-name
+    namespace: test-secret-namespace
+  additionalDeployments:
+    fake-controller-manager:
+      manager:
+        crdPattern: ".*"
```
alexander-demicev, line 102: Should `crdPattern` be removed since we can achieve the same with container arguments?

alexander-demicev:
PR looks good, just one small comment

willie-yao:
Looks like there's a problem with the E2E test output. Will be looking into this.

willie-yao:
The [hyphen in `--crd-pattern`](https://github.com/kubernetes-sigs/cluster-api-operator/compare/b6684d7966143e8b885bf6299d209928f85ee38a..151f8ee177b4e7981940b53cd1d49499902cc87f#diff-562d17ea1c2bd4a330a3731f8084721bbffe7aa3c0051459b8b8a7404899d196R105) messed up the test helm command, and since we're just testing if something can be set in the additionalDeployments rather than a specific field, I removed that field.

willie-yao:
Actually it looks like the error is coming from the recursivePrinter: https://github.com/willie-yao/cluster-api-operator/blob/25328eb918edd7eb4b2e79cf10ab73cf0ca4f9e1/hack/charts/cluster-api-operator/templates/infra.yaml#L1

I'm not sure how to adjust that function to allow array values to be passed inline. When specifying container customizations for additionalDeployments, using a values file works. Do we have to add support for setting this inline?

willie-yao:
I went ahead and removed the test case from `helm_test.go` due to the complicated nature of setting `additionalDeployments` via inline `--set`. If this is something we should support, I'll look into another way to scaffold out the additional customizations.

alexander-demicev:
@willie-yao We can look into setting `additionalDeployments` in tests in a follow-up PR

alexander-demicev (approved):
/approve

Fedosin (approved):
/lgtm

PR #561, opened 2024-07-02
📖 Fix manifestPatches field name

apricote:

The docs refer to this field as `resourcePatches` while its actually called [`manifestPatches`](https://github.com/kubernetes-sigs/cluster-api-operator/blob/628476cec06411357f07d514dfbf0444dad40c03/api/v1alpha2/provider_types.go#L75) in the code and in the CRD.

I also fixed the indentation in the examples for this field to make sense.


Danil-Grigorev (approved):
/lgtm
Thanks, this seems to be overlooked

alexander-demicev (approved):
/approve

Thanks!

PR #565, opened 2024-07-13
:sparkles: Handle config secret updates

tjamet:

As a cluster operator, I want to iterate my infrastructure provisioner
configuration and ensure the relevant cluster-api-providers are
provisioned as configured.

Currently, the GenericProvider Reconciler considers exclusively the
Providers spec and ignores the data pointed by the ConfigSecret field.

If one of those fields changes, the deployment of the provider is
unchanged.

For example, if an infrastructure is defined as

```yaml
apiVersion: operator.cluster.x-k8s.io/v1alpha2
kind: InfrastructureProvider
metadata:
  name: aws
  namespace: infrastructure-aws-system
spec:
  version: v2.5.2
  configSecret:
    name: aws-variables
  deployment:
    replicas: 1
---
apiVersion: v1
kind: Secret
metadata:
  name: aws-variables
  namespace: capi-config
type: Opaque
stringData:
  AWS_B64ENCODED_CREDENTIALS: "SOME_BASE_64_CREDS"
```

It is impossible to get the latest version of
`AWS_B64ENCODED_CREDENTIALS` without changing the provisioner version or
the deployments or verbosity level

This commit proposes to also take into consideration the content of the
configuration so, any change to it leads to an adjustment of the
deployment.

Currently, if a ConfigSecret is updated, it is not reflected automatically to all providers using it.
Add an optional reconciler to trigger the update of all providers using it.

Both combined ensures that any configuration update leads to an update of the provider deployment, with the least changes in behaviour possible


tjamet:
Digging into the failing test code, I don't understand how the proposed changes can influence the status.

the failing test (`TestCheckCAPIOpearatorAvailability` seems to test that deployments are working as expected.
It does create a deployment object running `nginx` with generateCAPIOperatorDeployment, update its status to consider it is running, and ensure that CheckDeploymentAvailability considers the deployment status as expected.

I can see the test [failing similarly in a dependabot PR](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-operator/556/pull-cluster-api-operator-test-main/1807747032031432704), leaning towards instability.

/retest

cmd/main.go
```diff
@@ -286,6 +291,24 @@ func setupReconcilers(mgr ctrl.Manager) {
 		setupLog.Error(err, "unable to create controller", "controller", "Healthcheck")
 		os.Exit(1)
 	}
+
+	if watchConfigSecretChanges {
```
Danil-Grigorev, line 295: I like the idea of watching secret changes, but I don’t like that it requires a separate controller. Hash of both object can be combined, and stored under the same annotation.

Controller can establish watch on secrets with object mapper, which will trigger reconcile for all providers where the secret is referenced.

tjamet: Thanks or the review.
I have updated the code to watch for secret changes and trigger a reconciliation of each Provider using them

Let me know what you think

internal/envtest/environment.go
```diff
@@ -125,6 +125,12 @@ func New(uncachedObjs ...client.Object) *Environment {
 		err = kerrors.NewAggregate([]error{err, env.Stop()})
 		panic(err)
 	}
+	u, _ := env.AddUser(envtest.User{Name: "debug", Groups: []string{"system:masters"}}, env.Config)
+
+	cfg, _ := u.KubeConfig()
+	fd, _ := os.Create("kubeconfig")
+	fd.Write(cfg)
+	defer fd.Close()
```
Danil-Grigorev, line 133: It seems this is a debug kubeconfig, can you please remove it? It may also fix the linter failure, but you can also run `make lint-fix` to verify.

internal/controller/genericprovider_controller.go
```diff
@@ -39,20 +42,28 @@ import (
...
 )
 
 func (r *GenericProviderReconciler) SetupWithManager(mgr ctrl.Manager, options controller.Options) error {
-	return ctrl.NewControllerManagedBy(mgr).
-		For(r.Provider).
-		WithOptions(options).
+	builder := ctrl.NewControllerManagedBy(mgr).
+		For(r.Provider)
+	if r.WatchConfigSecretChanges {
+		builder = builder.Watches(&corev1.Secret{}, &providerSecretMapper{
```
Danil-Grigorev, line 60: It may be better to use `handler.MapFunc`, similar to this, since using queue directly is “internal-ish” way of doing that.
```go
builder.Watch(
	source.Kind(mgr.GetCache(), &corev1.Secret{}),
	handler.EnqueueRequestsFromMapFunc(secretToProviders),
)
// where 
func secretToProviders(ctx context.Context, secret client.Object) []ctrl.Request
// loop over listConfigSecretUsers
```

Danil-Grigorev (commented):
Left some comments regarding CR usage, but overall looks good.

cc @furkatgofurov7 @alexander-demicev PTAL

internal/envtest/environment.go
tjamet: It is!
It shouldn't have been committed.
I will obviously remove it

internal/controller/genericprovider_controller.go
tjamet: I addressed your comment.
Apparently, `builder` does not have a `Watch` function but a `Watches` one which accepts a plain object as first parameter.
I though used `handler.EnqueueRequestsFromMapFunc` as suggested and adapted the rest of the code

Danil-Grigorev (approved):
/lgtm

Thank you @tjamet, this looks good to me.

Danil-Grigorev:
/lgtm
cc @furkatgofurov7 @alexander-demicev

internal/controller/secrets_to_providers_test.go
```diff
@@ -0,0 +1,121 @@
...
+						ConfigSecret: &operatorv1.SecretReference{
+							Name:      configSecretName,
+							Namespace: testNamespaceName,
+						},
+					},
+				},
+			},
+			&operatorv1.InfrastructureProvider{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "infra-provider-with-other-secret",
```
furkatgofurov7, line 83: to match the logic better IMO:
```suggestion
					Name:      "infra-provider-with-same-secret-different-namespace",
```

furkatgofurov7 (approved):
Changes overall looks good to me, thanks for your contribution @tjamet.

furkatgofurov7 (approved):
/approve

PR #638, opened 2024-11-25
⚠️ Make helm chart values extensible per component

rbjorklin:
When deploying clusters to AKS using Workload Identity [an annotation](https://learn.microsoft.com/en-us/azure/aks/workload-identity-deploy-cluster#create-a-kubernetes-service-account) is required on the service account and [a label](https://learn.microsoft.com/en-us/azure/aks/workload-identity-deploy-cluster#deploy-your-application) is required on the pod. By extending the chart to expose the [already existing functionality to patch provider manifests](https://cluster-api-operator.sigs.k8s.io/03_topics/02_configuration/04_patching-provider-manifests) this PR allows for setting up AKS clusters using Workload Identity.

Fixes #

### DIFF
--- a/hack/charts/cluster-api-operator/templates/addon.yaml
+++ b/hack/charts/cluster-api-operator/templates/addon.yaml
@@ -1,26 +1,8 @@
 # Addon provider
-{{- if .Values.addon }}
-{{- $addons := split ";" .Values.addon }}
-{{- $addonNamespace := "" }}
-{{- $addonName := "" }}
-{{- $addonVersion := "" }}
-{{- range $addon := $addons }}
-{{- $addonArgs := split ":" $addon }}
-{{- $addonArgsLen :=  len $addonArgs }}
-{{-  if eq $addonArgsLen 3 }}
-  {{- $addonNamespace = $addonArgs._0 }}
-  {{- $addonName = $addonArgs._1 }}
-  {{- $addonVersion = $addonArgs._2 }}
-{{-  else if eq $addonArgsLen 2 }}
-  {{- $addonNamespace = print $addonArgs._0 "-addon-system" }}
-  {{- $addonName = $addonArgs._0 }}
-  {{- $addonVersion = $addonArgs._1 }}
-{{-  else if eq $addonArgsLen 1 }}
-  {{- $addonNamespace = print $addonArgs._0 "-addon-system" }}
-  {{- $addonName = $addonArgs._0 }}
-{{- else }}
-  {{- fail "addon provider argument should have the following format helm:v1.0.0 or mynamespace:helm:v1.0.0" }}
-{{- end }}
+{{- range $name, $addon := $.Values.addon }}
+  {{- $addonNamespace := default ( printf "%s-%s" $name "addon-system" ) (get $addon "namespace") }}
+  {{- $addonName := $name }}
+  {{- $addonVersion := get $addon "version" }}
 ---
 apiVersion: v1
 kind: Namespace
@@ -52,5 +34,4 @@ spec:
 {{- if $.Values.secretNamespace }}
   secretNamespace: {{ $.Values.secretNamespace }}
 {{- end }}
-{{- end }} {{/* range $addon := $addons */}}
-{{- end }} {{/* if .Values.addon */}}
+{{- end }} {{/* range $name, $addon := .Values.addon */}}
--- a/hack/charts/cluster-api-operator/templates/addon.yaml
+++ b/hack/charts/cluster-api-operator/templates/addon.yaml
@@ -37,4 +37,21 @@ spec:
 {{- if $addon.manifestPatches }}
   manifestPatches: {{ toYaml $addon.manifestPatches | nindent 4 }}
 {{- end }}
+{{- if $addon.additionalManifests }}
+  additionalManifests:
+    name: {{ $addon.additionalManifests.name }}
+    {{- if $addon.additionalManifests.namespace }}
+    namespace: {{ $addon.additionalManifests.namespace }}
+    {{- end }} {{/* if $addon.additionalManifests.namespace */}}
+{{- end }}
+{{- if $addon.additionalManifests }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ $addon.additionalManifests.name }}
+  namespace: {{ default $addonNamespace $addon.additionalManifests.namespace }}
+data:
+  manifests: {{- toYaml $addon.additionalManifests.manifests | nindent 4 }}
+{{- end }}
 {{- end }} {{/* range $name, $addon := .Values.addon */}}
--- a/hack/charts/cluster-api-operator/templates/addon.yaml
+++ b/hack/charts/cluster-api-operator/templates/addon.yaml
@@ -34,4 +34,7 @@ spec:
 {{- if $.Values.secretNamespace }}
   secretNamespace: {{ $.Values.secretNamespace }}
 {{- end }}
+{{- if $addon.manifestPatches }}
+  manifestPatches: {{ toYaml $addon.manifestPatches | nindent 4 }}
+{{- end }}
 {{- end }} {{/* range $name, $addon := .Values.addon */}}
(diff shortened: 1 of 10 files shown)


furkatgofurov7:
@rbjorklin hey, would you be able to look into failing CI tests and fix them, so that we can start reviewing it ?

rbjorklin:
> @rbjorklin hey, would you be able to look into failing CI tests and fix them, so that we can start reviewing it ?

Fixed, please take a look!

hack/charts/cluster-api-operator/values.yaml
```diff
@@ -1,12 +1,30 @@
...
-core: ""
-bootstrap: ""
-controlPlane: ""
-infrastructure: ""
-ipam: ""
-addon: ""
+core: {}
+  # name: ""
+  # namespace: ""  # Optional
+  # version: ""  # Optional
```
rbjorklin, line 7: This is currently not a list because:
> [Note: Only one CoreProvider can be installed at the same time on a single cluster.](https://cluster-api-operator.sigs.k8s.io/03_topics/03_basic-cluster-api-provider-installation/01_installing-core-provider)

But maybe it should be a singleton list just to align it with the others?

rbjorklin:
@furkatgofurov7 given that this changes the structure of `values.yaml` should I introduce a [values.schema.json](https://helm.sh/docs/topics/charts/#schema-files) to make consumers of the chart aware of the change when they try rendering it?

hack/charts/cluster-api-operator/values.yaml
furkatgofurov7: I think it can stay as is, can't think of the point it being a list tbh

furkatgofurov7:
> @furkatgofurov7 given that this changes the structure of `values.yaml` should I introduce a [values.schema.json](https://helm.sh/docs/topics/charts/#schema-files) to make consumers of the chart aware of the change when they try rendering it?

@rbjorklin good idea, it seems to be useful when using `helm lint`?

rbjorklin:
@furkatgofurov7 I have just pushed a minimal `values.schema.json` to catch the breaking changes to the `values.yaml` file introduced in this PR. When trying to template the chart a user will be faced with something like this:

```console
❯ helm template test hack/charts/cluster-api-operator
Error: values don't meet the specifications of the schema(s) in the following chart(s):
cluster-api-operator:
- ipam: Array must have at least 1 items
- core: name is required
- bootstrap: Array must have at least 1 items
- controlPlane: Array must have at least 1 items
- infrastructure: Array must have at least 1 items
- addon: Array must have at least 1 items
```

hack/charts/cluster-api-operator/values.schema.json
```diff
@@ -0,0 +1,62 @@
...
+      "additionalProperties": false
+    }
+  },
+  "properties": {
+    "core": {
+      "$ref": "#/definitions/nameNamespaceVersion"
+    },
+    "bootstrap": {
+      "type": "array",
+      "minItems": 1,
```
rbjorklin, line 28: Setting `minItems: 1` might break the tests but how useful is this chart if you do not specify all of the providers? Are any of these actually optional?

rbjorklin: This did indeed break the tests. Removed.

mjura (approved):
LGTM

hack/charts/cluster-api-operator/values.yaml
```diff
@@ -1,12 +1,30 @@
...
+bootstrap: []
```
afarbos, line 8: I personally like the fact a string and not a list since it's really hard with helm to override values in list.

If you move with change please consider a dictionary instead, something like:
```
bootstrap:
   <name>:
     namespace: "" # optional
     version: ""        # optional
     enabled: true   # optional
```

rbjorklin: That’s a valid argument and I’m not opposed to it. If this gets another vote from someone I’ll rework the PR.
Maybe @furkatgofurov7 can chime in?

furkatgofurov7: I am fine with it as well

test/e2e/helm_test.go
```diff
@@ -184,14 +191,14 @@ var _ = Describe("Create a proper set of manifests when using helm charts", func
...
-			"configSecret.namespace": "test-secret-namespace",
-			"core":                   "cluster-api",
-			"controlPlane":           "kubeadm",
-			"bootstrap":              "kubeadm",
-			"infrastructure":         "docker",
-			"ipam":                   "in-cluster",
-			"addon":                  "helm",
+			"configSecret.name":          "test-secret-name",
+			"configSecret.namespace":     "test-secret-namespace",
+			"core.cluster-api.name":      "null",
```
rbjorklin, line 196: The `name` field doesn't actually exist and is never used for anything. This is just here to make the types line up, if anyone has a better solution I'm all ears.

rbjorklin:
@furkatgofurov7 @afarbos reworked and tests are passing. Please take a look! :pray:

test/e2e/helm_test.go
afarbos: Yes, I am not sure if that make sense, this is just to set a key. that's why I was suggesting also adding `enabled`. This would solve problem with empty dictionaries values

```suggestion
			"core.cluster-api.enabled":      "true",
```
 looking here https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set

you might be able to do

```suggestion
			"core":      "cluster-api",
```
or 
```suggestion
			"core.cluster-api":      "null",
```

I do like `enabled` more since it allow to easily turn off a provider.

rbjorklin: Just pushed the `enabled` version.

rbjorklin:
I've addressed the review comment by @afarbos. I'd like to believe that this is in merge:able shape now!

hack/charts/cluster-api-operator/values.yaml
```diff
@@ -1,12 +1,30 @@
 ---
 # ---
 # Cluster API provider options
-core: ""
-bootstrap: ""
-controlPlane: ""
-infrastructure: ""
-ipam: ""
-addon: ""
```
furkatgofurov7, line 9: So, now options have been updated to use structured objects for providers instead of plain strings. Could it be potentially a breaking change since existing deployments rely on the old string format and need to be updated to the new format?

afarbos: It is, this would be a major upgrade for the helm chart.

hack/charts/cluster-api-operator/templates/addon.yaml
```diff
@@ -1,26 +1,8 @@
...
-{{-  else if eq $addonArgsLen 1 }}
-  {{- $addonNamespace = print $addonArgs._0 "-addon-system" }}
-  {{- $addonName = $addonArgs._0 }}
-{{- else }}
-  {{- fail "addon provider argument should have the following format helm:v1.0.0 or mynamespace:helm:v1.0.0" }}
-{{- end }}
+{{- range $name, $addon := $.Values.addon }}
+  {{- $addonNamespace := default ( printf "%s-%s" $name "addon-system" ) (get $addon "namespace") }}
+  {{- $addonName := $name }}
+  {{- $addonVersion := get $addon "version" }}
```
afarbos, line 5:
```suggestion
  {{- $addonVersion := get $addon "version" }}
  {{- $addonEnabled := default (get $addon "enabled") true }}
  {{- if not $addonEnabled }}
  {{- continue }}
  {{- end }}
```
I would have use enabled, no?

rbjorklin: I figured `range`ing over the the dict was sufficient. If the dict is empty no manifests will be created. I can add the `enabled` check if you prefer though.

Sidenote: Helm does not actually have a `continue` keyword.

alexander-demicev (approved):
/approve

furkatgofurov7 (commented):
Thanks @rbjorklin for patience and pushing this to the finish line.

/retitle ⚠️ Make helm chart values extensible per component
/lgtm

furkatgofurov7:
@rbjorklin hey, I noticed we forgot to have some docs around this breaking change. Would you be willing to contribute to document it, which would greatly help existing/new users facing challenges with the new values schema.

PR #650, opened 2024-12-13
✨ adding fetchConfig to InfrastructureProvider template in helmchart

ts-mini:
**What this PR does / why we need it**: This adds [fetchConfig](https://doc.crds.dev/github.com/kubernetes-sigs/cluster-api-operator/operator.cluster.x-k8s.io/InfrastructureProvider/v1alpha2) to the `InfrastructureProvider`.  

This can be used like this:
```
infrastructure: "oci"
manager:
  featureGates:
    oci:
      MachinePool: true
      OKE: true
fetchConfig:
  oci:
    url: https://github.com/ts-mini/cluster-api-provider-oci/releases
```

The attempt was to follow the existing patterns of per-provider config, allowing users to set fetchConfigs per provider if needed.  Happy to adjust as needed!

### DIFF
--- a/hack/charts/cluster-api-operator/templates/infra.yaml
+++ b/hack/charts/cluster-api-operator/templates/infra.yaml
@@ -59,6 +59,16 @@ spec:
 {{- end }}
 {{- end }}
 {{- end }}
+{{- if and (kindIs "map" $.Values.fetchConfig) (hasKey $.Values.fetchConfig $infrastructureName) }}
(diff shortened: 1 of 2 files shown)


alexander-demicev (approved):
thanks for the PR, it would be good to also add this functionality to other types of providers
/approve

furkatgofurov7 (commented):
/lgtm

PR #688, opened 2025-01-25
✨ Add IPAM provider support to Cluster API Operator Helm chart

rgeraskin:
**What this PR does / why we need it**:

This PR introduces a Helm chart template for deploying IPAMProvider.

PR #549, opened 2024-06-03
🌱  Add more cases to helm test

willie-yao:
**What this PR does / why we need it**:
This PR adds a few test cases that would've caught the issues found in #532. It also switches the yaml comparison from `Equal` to `MatchYAML` which prints a much more informative message for failures.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #692, opened 2025-01-28
🐛 Fix helm chart e2e tests: MatchYAML to Equal

rgeraskin:
**What this PR does**

In this PR, I’ve reverted MatchYAML to Equal, fixed all the broken tests. Also, I discovered and fixed a bug in the core provider specification condition.

**Why we need it**:

While working on PR [#688,](https://github.com/kubernetes-sigs/cluster-api-operator/pull/688) I’ve observed that a significant number of Helm chart e2e tests are false positives. This is because the generated YAML files in the test/e2e/resources directory do not accurately reflect the expected results from the test cases.

I believe this issue arises because `MatchYAML` is being used instead of `Equal` ([#549](https://github.com/kubernetes-sigs/cluster-api-operator/pull/549)). `MatchYAML` only compares the first YAML document within a file, as demonstrated in a quick example.

```
package main_test

import (
	"testing"

	. "github.com/onsi/ginkgo/v2"
	. "github.com/onsi/gomega"
)

func TestGomega(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Gomega Suite")
}

var _ = Describe("MatchYAML compares only the first document", func() {
	It("equals", func() {
		yaml1 := `---
hello: world
---
actually: do not care about this document
`
		yaml2 := `---
hello: world
---
even if it is not a valid yaml
`
		Expect(yaml1).To(MatchYAML(yaml2))
	})
})
```

### DIFF
--- a/hack/charts/cluster-api-operator/templates/core.yaml
+++ b/hack/charts/cluster-api-operator/templates/core.yaml
@@ -38,7 +38,7 @@ metadata:
     "helm.sh/hook": "post-install,post-upgrade"
     "helm.sh/hook-weight": "2"
     "argocd.argoproj.io/sync-wave": "2"
-{{- if or $coreVersion $.Values.configSecret.name }}
+{{- if or $coreVersion $.Values.configSecret.name $.Values.manager }}
 spec:
 {{- end}}
 {{- if $coreVersion }}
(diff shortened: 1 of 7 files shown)


roman-geraskin-m10:
Thanks for the fast review 🙂

rgeraskin:
Thanks for the fast review 🙂

PR #703, opened 2025-01-30
:sparkles: Handle Kubernetes events for waiting CoreProvider in preflight check

dmvolod:
This PR allows you to avoid from polling in reconcile loop every 30 seconds, and move to catching events from the CoreProvider object when Ready condition become True. We should avoid re-querying at intervals where possible.

Also, configSecret request is optimized if option is enabled. Since a cluster can have a large number of secret objects, it is a very expensive procedure to go through all the providers each time secrets are catches without an index. The number of provider objects in the cluster is small enough that the index on the name/namespace fields will not be very large.

Fixes #

### DIFF
--- a/internal/controller/coreprovider_to_providers.go
+++ b/internal/controller/coreprovider_to_providers.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"fmt"
+
+	operatorv1 "sigs.k8s.io/cluster-api-operator/api/v1alpha2"
+	"sigs.k8s.io/cluster-api-operator/internal/controller/genericprovider"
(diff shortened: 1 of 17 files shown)


internal/controller/coreprovider_to_providers.go
```diff
@@ -0,0 +1,69 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
furkatgofurov7, line 2: nit:

```suggestion
Copyright 2025 The Kubernetes Authors.
```

internal/controller/coreprovider_to_providers_test.go
```diff
@@ -0,0 +1,143 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
furkatgofurov7, line 2: same

internal/envtest/environment.go
```diff
@@ -298,26 +298,13 @@ func (e *Environment) CreateNamespace(ctx context.Context, generateName string)
...
-
-	// Create the namespace if it doesn't exist
 	newNamespace := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: namespace,
 		},
 	}
 
-	if err := e.Client.Create(ctx, newNamespace); err != nil {
+	if err := e.Client.Create(ctx, newNamespace); client.IgnoreAlreadyExists(err) != nil {
```
furkatgofurov7, line 307: 👍🏼

furkatgofurov7 (commented):
Nice work, few non-blocking nits inline + lgtm pending linter fix:

internal/controller/coreprovider_to_providers_test.go
dmvolod: Yes, sure, fixed.
In this case, we could probably change a `goheader` linter settings and replace `license-year` header with `MOD-YEAR` value and golangci-lint will take case itself.

furkatgofurov7 (approved):
/approve

Danil-Grigorev (commented):
/lgtm

PR #706, opened 2025-01-30
✨ Allow accepting self-hosted gitlab installations

### LINKED ISSUE #705: Cluster API fetchConfig accepting self-hosted gitlab installations
**User Story**

As a operator I would like to deployed a cluster-api-providers from self-managed gitlab installations that dont match `gitlab.`

**Detailed Description**

Currently, the [code](https://github.com/kubernetes-sigs/cluster-api-operator/blob/main/util/util.go#L33-L37) only accepts gitlab installations that start with `gitlab.` and github installations that end in `github.com`.  The goal would be to allow `fetchConfig` the ability to accept a url that doesn't match this explicit pattern as both tools offer self hosting capabilities.

/kind feature

ts-mini:
**What this PR does / why we need it**: The current implementation of the gitlab check uses HasPrefix - which doesn't allow for a gitlab whose domain could be different (gitlab-example.domain.com)

Fixes #705

### DIFF
--- a/internal/controller/phases_test.go
+++ b/internal/controller/phases_test.go
@@ -642,6 +642,10 @@ func TestRepositoryFactory(t *testing.T) {
 			fetchURL: "https://gitlab.example.org/api/v4/projects/group%2Fproject/packages/generic/cluster-api-proviver-aws/v1.4.1/path",
 		},
 		{
+			name:     "gitlab hyphenated repo",
+			fetchURL: "https://gitlab-test.example.org/api/v4/projects/group%2Fproject/packages/generic/cluster-api-proviver-aws/v1.4.1/path",
+		},
+		{
 			name:          "unsupported url",
 			fetchURL:      "https://unsupported.xyz/kubernetes-sigs/cluster-api-provider-aws/releases/v1.4.1/infrastructure-components.yaml",
 			expectedError: true,
--- a/util/util.go
+++ b/util/util.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"regexp"
 	"strings"
 
 	operatorv1 "sigs.k8s.io/cluster-api-operator/api/v1alpha2"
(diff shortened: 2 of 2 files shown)


internal/controller/phases_test.go
```diff
@@ -641,6 +641,10 @@ func TestRepositoryFactory(t *testing.T) {
 			name:     "gitlab repo",
 			fetchURL: "https://gitlab.example.org/api/v4/projects/group%2Fproject/packages/generic/cluster-api-proviver-aws/v1.4.1/path",
 		},
+		{
+			name:     "gitlab hyphenated repo",
+			fetchURL: "https://gitlab-test.example.org/api/v4/projects/group%2Fproject/packages/generic/cluster-api-proviver-aws/v1.4.1/path",
```
dmvolod, line 646: This is a pretty good idea, but it is a special case.
For example, our private GitLab instance called git.<our domain> can't be changed and no `gitlab` prefix or postfix in name.
It would be nice to have some sort of option which can say that it's a GitLab instance if there is not possible to define it form the FQDN.

But seems to community should define is if suitable approach or not.

ts-mini: generally agree - the change to the code to support what i consider 'the more correct approach' wasn't within my allotted time as  it seems like it should live in the [providerConfig](https://github.com/kubernetes-sigs/cluster-api-operator/blob/tsmini.support-any-gitlab/internal/controller/manifests_downloader.go#L82-L83) of `Provider type` - which would be large change - perhaps percolating to all providers.

while this does highlight a special use case - i'm wondering if its _generic enough_

alexander-demicev (approved):
/approve

internal/controller/phases_test.go
alexander-demicev: We can find a more flexible approach in a follow-up PR

dmvolod: Ok, make sense.

furkatgofurov7:
/retitle ✨ Allow accepting self-hosted gitlab installations

PR #724, opened 2025-02-10
:bug: ConfigMap OCI Image Label Error

### LINKED ISSUE #720: OCI Artifact URL ConfigMap Label Length
The following code block is causing issues when providing an OCI Artifact URL that is longer than allowed in labels:

https://github.com/kubernetes-sigs/cluster-api-operator/blob/8647abf90930d91c835720cd7ed00554aa0d5baa/internal/controller/manifests_downloader.go#L295-L299

I am writing up a PR to fix this, will add it as an `annotation` instead of a `label`, but am looking for feedback on what the preferred approach is.

Nalum:
The current code puts the OCI Artifact URL into a label without consideration for the allowed length of label values, this PR will set the Source label to `oci` and add the OCI URL as an annotation on the ConfigMap.

Fixes #720

### DIFF
--- a/internal/controller/manifests_downloader.go
+++ b/internal/controller/manifests_downloader.go
@@ -21,7 +21,6 @@ import (
 	"compress/gzip"
 	"context"
 	"fmt"
-	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


furkatgofurov7 (approved):
/approve

cc @Danil-Grigorev

Danil-Grigorev (approved):
/lgtm
Thank you!

PR #725, opened 2025-02-12
:seedling: Add alias to GenericProviderReconciler type

dmvolod:
This fix allows to import GenericProviderReconciler and embed it into the third-party golang code for testing and utilizing standalone without fork. 
Also moves some useful labels to the API and make it public.

Fixes issue in #727


internal/controller/manifests_downloader.go
```diff
@@ -38,9 +38,6 @@ import (
 )
 
 const (
-	configMapVersionLabel     = "provider.cluster.x-k8s.io/version"
```
furkatgofurov7, line 41: what is the benefit of moving these to API package? AFAICT, they are used in internal package only

dmvolod: Yes, this is good question, but according to test data below, it seems to possible to have any ConfigMap name with following labels to be configured in air-gap mode, to avoid have name equals to version.

This would be an option for configuration. Seems to need to add this option to documentation.
What do you think about it?

https://github.com/kubernetes-sigs/cluster-api-operator/blob/6b72fcd6c67e60a1f80ef1f367f065d4adec7019/test/e2e/resources/core-cluster-api-v1.7.7.yaml#L13358-L13360

furkatgofurov7:
can you run` make-generate` to fix api-diff issue?

dmvolod:
> can you run` make-generate` to fix api-diff issue?

Yes, sure, @furkatgofurov7 
`ConfigMapVersionLabelName` name was released in v0.16.0 and seems to shouldn't be renamed, right?

furkatgofurov7:
> > can you run` make-generate` to fix api-diff issue?
> 
> Yes, sure, @furkatgofurov7 `ConfigMapVersionLabelName` name was released in v0.16.0 and seems to shouldn't be renamed, right?

Not sure I am following, but changes look good to me

/approve

mjura (commented):
LGTM

alexander-demicev (approved):
/lgtm

Issue #562, opened 2024-07-06
Automatic deletion of all providers


dtzar:
**What steps did you take and what happened:**
Install helm chart using ArgoCD and it starts to do the install and then for some unknown reason the capi-operator-system deletes everything.

**What did you expect to happen:**
capi-operator-system doesn't initiate deletion

```
I0705 19:26:02.047065       1 phases.go:501] "Installing provider" controller="infrastructureprovider" controllerGroup="operator.cluster.x-k8s.io" controllerKind="InfrastructureProvider" InfrastructureProvider="azure-infrastructure-system/azure" namespace="azure-infrastructure-system" name="azure" reconcileID="f522a400-3574-49f2-a2f9-f5cd647d1353"
I0705 19:26:08.481831       1 genericprovider_controller.go:62] "Reconciling provider" controller="bootstrapprovider" controllerGroup="operator.cluster.x-k8s.io" controllerKind="BootstrapProvider" BootstrapProvider="capi-kubeadm-bootstrap-system/kubeadm" namespace="capi-kubeadm-bootstrap-system" name="kubeadm" reconcileID="55d807bc-19a8-4fd9-a894-9586199f2dda"
I0705 19:26:08.482636       1 genericprovider_controller.go:190] "Deleting provider resources" controller="bootstrapprovider" controllerGroup="operator.cluster.x-k8s.io" controllerKind="BootstrapProvider" BootstrapProvider="capi-kubeadm-bootstrap-system/kubeadm" namespace="capi-kubeadm-bootstrap-system" name="kubeadm" reconcileID="55d807bc-19a8-4fd9-a894-9586199f2dda"
I0705 19:26:08.482750       1 phases.go:547] "Deleting provider" controller="bootstrapprovider" controllerGroup="operator.cluster.x-k8s.io" controllerKind="BootstrapProvider" BootstrapProvider="capi-kubeadm-bootstrap-system/kubeadm" namespace="capi-kubeadm-bootstrap-system" name="kubeadm" reconcileID="55d807bc-19a8-4fd9-a894-9586199f2dda"
[...]

PR #751, opened 2025-03-14
🐛  Improve deployment with Argocd by allowing to remove helm hooks 

afarbos:

Allow to disable have helm hooks. This help with argocd resource management (cannot autosync, no proper tracking of resources on deletion) and avoid deleting all hook resources each time we need to sync. This can be scary if the un-install does not work as expected and stuck the app state with no provider running.

Argocd automatically convert hooks from helm and there is no way to avoid that once those are set the resource will be considered deployed by a hook.
For reference here is the conversion: https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks
By enabling the removal of helm hooks, argocd will still install the resource in the right order using the phase only ( https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/#how-do-i-configure-waves )

Fixes https://github.com/kubernetes-sigs/cluster-api-operator/issues/562

### DIFF
--- a/hack/charts/cluster-api-operator/templates/addon.yaml
+++ b/hack/charts/cluster-api-operator/templates/addon.yaml
@@ -26,8 +26,10 @@ apiVersion: v1
 kind: Namespace
 metadata:
   annotations:
+    {{- if $.Values.enableHelmHook }}
     "helm.sh/hook": "post-install,post-upgrade"
     "helm.sh/hook-weight": "1"
+    {{- end }}
     "argocd.argoproj.io/sync-wave": "1"
   name: {{ $addonNamespace }}
 ---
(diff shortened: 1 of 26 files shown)


furkatgofurov7:
/ok-to-test

cc @dtzar would it be possible to test this patch before we take it in?

afarbos:
til `make test-e2e-local`, make it way simpler to debug those!
🤞 test should pass

alexander-demicev (approved):
/approve

please, squash commits

furkatgofurov7:
LGTM pending squash

afarbos:
Squashed. see  https://github.com/kubernetes-sigs/cluster-api-operator/compare/2e05fb1ac05157cee318bcc973e57d731565bc2d..5ea7523b87292e595e7c17fa49aec2bdf804579f for diff

furkatgofurov7 (commented):
Thanks for the PR @afarbos!

/lgtm

PR #779, opened 2025-04-07
:seedling: Bump golangci-lint to v2

anmazzotti:

This bumps the golangci-lint version to v2.
Should address and deprecate #766

The new config was [migrated](https://golangci-lint.run/product/migration-guide/).
I replicated the comments in the migrated version.

Also note that I added some exclusion rules:

```
      - linters:
          - staticcheck
        text: 'ST1016: methods on the same type should have the same receiver name'
        path: api/(.+)\.go$
      - linters:
          - staticcheck
        text: 'ST1003: should not use underscores in Go names;'
        path: api/(.+)\.go$      
      - linters:
          - staticcheck
        text: 'QF1008: could remove embedded field'
      - linters:
          - staticcheck
        text: 'ST1000: at least one file in a package should have a package comment'
```

`ST1016` and `ST1003` only apply to `api/` files, for convenience with the conversions code readability.
`ST1000` I do not find that useful personally.

 `QF1008` I do like, but there were too many instances to fix and I don't know what the consensus is, so I excluded it before making too many changes to the codebase.

Fixes #

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -21,13 +21,13 @@ jobs:
         with:
           go-version: ${{ steps.vars.outputs.go_version }}
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@55c2c1448f86e01eaae002a5a3a9624417608d84 # tag=v6.5.2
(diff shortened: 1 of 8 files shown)


furkatgofurov7 (approved):
Thanks @anmazzotti 

/approve
/lgtm

PR #780, opened 2025-04-08
:sparkles: Make configSecret configurable per component

rbjorklin:
This builds on the extensibility introduced in #638.

* This makes the connection between the `configSecret` and the component/provider using said config clearer.
* Reduces duplication of the `configSecret` logic across templates by moving it into a helper.
* Adds the ability to configure `fetchConfig`.

This has been implemented in such a way that it is backwards compatible and will default to the existing behaviour. If you want to use the new behaviour you would have to do something like this:

```yaml
core:
  cluster-api:
    version: v1.9.6
    configSecret:
      name: my-custom-core-config
      namespace: capi-system

infrastructure:
  azure:
    namespace: capz-system
    version: v1.19.1
    configSecret:
      name: azure-variables
      namespace: capz-system
```
while also removing the `configSecret` key at the root level of the `values.yaml` file.

Fixes #

### DIFF
--- a/hack/charts/cluster-api-operator/templates/_helpers.tpl
+++ b/hack/charts/cluster-api-operator/templates/_helpers.tpl
@@ -22,3 +22,13 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{- define "capi-operator.configSecret" -}}
+{{- $ := .ROOT -}}
+{{- $arg := .ARGUMENT -}}
+configSecret:
+  name: {{ default (($arg).configSecret).name (($.Values).configSecret).name }}
+  {{- if (default (($arg).configSecret).namespace (($.Values).configSecret).namespace) }}
+  namespace: {{ default (($arg).configSecret).namespace (($.Values).configSecret).namespace }}
+  {{- end }}
+{{- end -}}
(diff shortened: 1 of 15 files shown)


sathieu:
@rbjorklin Thanks for this. Following the same idea, can you add `fetchConfig`, `manager.featureGates`, `additionalDeployments`, `manifestPatches`, and `additionalManifests` configurable per provider?

rbjorklin:
> @rbjorklin Thanks for this. Following the same idea, can you add `fetchConfig`, `manager.featureGates`, `additionalDeployments`, `manifestPatches`, and `additionalManifests` configurable per provider?

`manifestPatches` and `additionalManifests` were already implemented in the referenced PR #638. I'll see if I can get the others added though.

bsctl:
@rbjorklin @sathieu what's about to merge these PR #780 and leave open other PRs for the remaining `fetchConfig`, `manager.featureGates`, `additionalDeployments` ?

sathieu:
@bsctl I've moved to another way of managing providers ([here](https://gitlab.com/kubitus-project/apps/cluster-api-operator/-/blob/main/charts/cluster-api-operator/values.yaml?ref_type=heads#L52)) and I'm no longer interrested in this PR.

rbjorklin:
Just rebased and added `fetchConfig`. Is there a maintainer that could review this?

rbjorklin:
Alright, I think I've got the tests sorted now. This just needs a review.

PR #812, opened 2025-05-07
🌱 Pass global concurrency number to manager

Danil-Grigorev:

This PR slightly improves memory footprint of the operator by implementing minor adjustments to the code:
- Skip components.yaml caching in upgrade scenario for all other providers except upgraded. Only metadata.yaml is important there.
- Pass concurrency number to the manager instance, allowing max X concurrent reconciles across different providers.
- Increases memory limits to 300Mi from 150Mi, since in current setup azure provider alone is able to bump memory usage to median of ~130Mi.

Fixes #


furkatgofurov7 (commented):
In general improvements make sense to me, thanks for the PR. 

/lgtm

alexander-demicev (approved):
/approve

PR #813, opened 2025-05-12
Make namespace creation optional and configurable per component

fariaass:
🌱 (:seedling:, minor or other)

This PR adds the option to choose whether to create namespaces together the providers or not while helm chart installing. It makes the option configurable per component.


fariaass:
Hey guys! How are you doing?
Do you have any updates about this pls?

furkatgofurov7 (commented):
Having a self-contained chart to me looks more favorable where the chart manages everything needed, however, I do understand also the use cases where it should be optional and that is useful, since at the end of the day the default behaviour is staying the same (we always create ns unless it is explicitly set not to do so) after this change, so I am fine with the change. 

/lgtm

PR #814, opened 2025-05-15
 🐛 Fix OCI source parsing issues

neoaggelos:

Previously, an OCI source "registry:port/image" would consider ":port/image" as the tag

Add unit tests to verify change works as desired, and prevent regressions in the future

Did not create an issue to go with this PR, happy to do so if requested by the maintainers


furkatgofurov7 (commented):
/lgtm

Issue #534, opened 2024-05-24
Helm Chart CRDs Placement Causes Flaky Installations


kahirokunn:
**What steps did you take and what happened:**
Attempting to install the Helm chart for `cluster-api-operator` results in a flaky installation process. The `CRDs` are located in the `templates/operator-components.yaml` file instead of the `crds` directory.

```sh
helm upgrade --install cluster-api-operator-docker cluster-api-operator/cluster-api-operator --version 0.10.1 -n cluster-api-operator-docker --create-namespace --values ./cluster-api-operator-values.yaml
Release "cluster-api-operator-docker" does not exist. Installing it now.
Error: failed post-install: warning: Hook post-install cluster-api-operator/templates/core-conditions.yaml failed: 1 error occurred:
        * Internal error occurred: failed calling webhook "vcoreprovider.kb.io": failed to call webhook: Post "https://capi-operator-webhook-service.cluster-api-operator-docker.svc:443/mutate-operator-cluster-x-k8s-io-v1alpha2-coreprovider?timeout=10s": no endpoints available for service "capi-operator-webhook-service"
```

**What did you expect to happen:**
The Helm install should complete successfully without errors.

**Anything else you would like to add:**
It is possible to separate the `CustomResourceDefinitions` from the other components using commands like the following:

```sh
cat operator-components.yaml | yq 'select(.kind != "CustomResourceDefinition")'
cat operator-components.yaml | yq 'select(.kind == "CustomResourceDefinition")'
```

Can we consider reorganizing the files accordingly?

**Environment:**

- Cluster-api-operator version:
- Cluster-api version:
- Minikube/KIND version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

/kind bug
[One or more /area labels. See https://github.com/kubernetes-sigs/cluster-api-operator/labels?q=area for the list of labels]

---

Feel free to fill in the specific versions and any additional environment details before submitting the issue.

kahirokunn:
We have verified that cluster-api-operator cannot be successfully installed by `helm upgrade --install` .
The helm install works.

kahirokunn:
It won't work if you move it as it is because CRDs uses the go template. 😢

kahirokunn:
Oh, Thx.

kahirokunn:
Okay👌

PR #832, opened 2025-06-19
fix: split provider CRs from operator deployment

kahirokunn:
This commit splits the cluster-api-operator Helm chart into two separate charts to resolve flaky installations caused by webhook validation timing issues.

Problem:
- Provider Custom Resources (CoreProvider, BootstrapProvider, etc.) were applied at the same time as the operator deployment.
- The webhook service was not yet ready, leading to validation errors: "no endpoints available for service 'capi-operator-webhook-service'".

Solution:
- Create two charts:
  1. cluster-api-operator: contains only the operator deployment and its resources.
  2. cluster-api-operator-providers: contains all provider Custom Resources.
- Installing the operator first allows the webhook to start before provider CRs are applied.

Installation now requires:
1. Install operator: helm install capi-operator capi-operator/cluster-api-operator \ --create-namespace -n capi-operator-system --wait --timeout 90s
2. Install providers: helm install capi-providers \ capi-operator/cluster-api-operator-providers \ -n capi-operator-system \ --set infrastructure.docker.enabled=true \ --set cert-manager.enabled=true \ --set configSecret.name=${CREDENTIALS_SECRET_NAME} \ --set configSecret.namespace=${CREDENTIALS_SECRET_NAMESPACE}

Fixes: #534

Fixes #


kahirokunn:
Hi @dtzar and @furkatgofurov7,

When you have a moment, could you please review ?

This update splits the chart into operator and provider components, ensuring the operator’s webhook is fully ready before provider CRs are applied. As a result, the `no endpoints available for service "capi-operator-webhook-service"` validation error **has been resolved**.

Thank you for your time and feedback.

Best regards,
@kahirokunn

Danil-Grigorev:
Hey @kahirokunn, can you PTAL at failing tests? Thanks

Danil-Grigorev:
It makes sense in general, but providers chart seem to be a better fit as an operator chart dependency. This, depending on the configuration may make this change non-breaking

kahirokunn:
@Danil-Grigorev I tried changing it, but how does it look? 👀

hack/charts/cluster-api-operator-providers/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+name: cluster-api-operator-providers
+description: Cluster API Provider Custom Resources
+type: application
+version: 0.0.0
+appVersion: "0.0.0"
+dependencies:
+  - name: cluster-api-operator
+    repository: file://../cluster-api-operator
```
Danil-Grigorev, line 9: It probably won’t work outside of locally built chart. On the second thought, having charts separated would be simpler from release perspective

kahirokunn: Thx. I did it.

rbjorklin:
@kahirokunn I see that you're also implementing `configSecret` support in a similar way I did in #780. If you wouldn't mind adding `fetchConfig` support in a similiar way to [this](https://github.com/kubernetes-sigs/cluster-api-operator/pull/780/commits/4fdac3e68834a5a279bdaeb914d659f333dc6ebd) I would be happy to close my PR and endorse yours instead!

hack/charts/cluster-api-operator/values.schema.json
```diff
@@ -2,46 +2,150 @@
   "$schema": "http://json-schema.org/draft-07/schema#",
```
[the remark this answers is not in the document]
kahirokunn: Thank you. I fixed it.

kahirokunn:
@rbjorklin Thank you! I've cherry-picked the fetchConfig implementation from your PR #780 and integrated it with the configSecret support.I appreciate you offering to consolidate - your fetchConfig approach works well with what I had. Thanks for the collaboration!

PR #852, opened 2025-07-14
✨ Expose clusterctl provider dependent logic for custom provider building

Danil-Grigorev:

CAPI Operator reconciles essentially 2 provider types: Core and non-Core providers. There are differences in `GetType()` method and how each provider translates to the path to the component files in the release payload, but these details are abstracted away behind `GenericProvider` interface.

This change exposes inner methods used to determine provider type and provide a `clusterctl` equivalent for it via a set of injection points, available as a public API through `alias.go`

Fixes #


alexander-demicev (approved):

anmazzotti (approved):
/lgtm

PR #855, opened 2025-07-17
🐛 Fix preflight check for generic provider overlap with core

Danil-Grigorev:

In a scenario when a homogeneous provider type is used, core provider preflight check reports a false negative failure, due to encountering more than one core provider in the list.

To fix this, we need to compare provider type for both current and found at all times.

Fixes #


furkatgofurov7 (commented):
/lgtm

Issue #858, opened 2025-07-23
[OCI] Wrong contolplane components file pattern


mjura:
**What steps did you take and what happened:**
Try to use OCI fetcConfig to deploy kubeadm controlplane

```
apiVersion: v1
kind: Namespace
metadata:
  name: capi-kubeadm-control-plane-system
---
apiVersion: turtles-capi.cattle.io/v1alpha1
kind: CAPIProvider
metadata:
  name: kubeadm-control-plane
  namespace: capi-kubeadm-control-plane-system
spec:
  name: kubeadm
  type: controlplane
  version: v1.10.2
  fetchConfig:
    oci: registry.rancher.com/rancher/cluster-api-controller-components:v1.10.2
```

**What did you expect to happen:**
Provider kubeadm-control-plane will be provisioned:
```
kubectl get capiproviders.turtles-capi.cattle.io -A

NAMESPACE                           NAME                    TYPE             PROVIDERNAME   INSTALLEDVERSION   PHASE
capi-kubeadm-control-plane-system   kubeadm-control-plane   controlplane     kubeadm        v1.10.2            Ready
```

**Anything else you would like to add:**

Currently kubeadm and rke2 provider are generating controlPlane components files with wrong name:
`control-plane-components.yaml`

Error message during provisioning:

```
...
  - lastTransitionTime: "2025-07-23T06:53:55Z"
    message: collected artifact needs to provide components as controlplane-kubeadm-v1.10.2-components.yaml
      or controlplane-components.yaml or components.yaml file
    reason: ComponentsFetchError
...
```

**Environment:**

- Cluster-api-operator version:
- Cluster-api version:
- Minikube/KIND version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

/kind bug
[One or more /area label. See https://github.com/kubernetes-sigs/cluster-api-operator/labels?q=area for the list of labels]

Danil-Grigorev:
Closed by #859

PR #859, opened 2025-07-23
🐛 [OCI] Normalize provider type name when control plane

mjura:
Normalize provider type name for controlPlane, both kubeadm and rke2 contolPlane components have name control-plane-components.yaml

Issue: https://github.com/kubernetes-sigs/cluster-api-operator/issues/858

Fixes #

### DIFF
--- a/api/v1alpha2/controlplaneprovider_wrapper.go
+++ b/api/v1alpha2/controlplaneprovider_wrapper.go
@@ -47,7 +47,7 @@ func (c *ControlPlaneProvider) SetStatus(in ProviderStatus) {
 }
 
 func (c *ControlPlaneProvider) GetType() string {
-	return "controlplane"
+	return "control-plane"
 }
 
 func (c *ControlPlaneProvider) ProviderName() string {


Danil-Grigorev (commented):
Thanks for the fix. Possibly this needs to be charged in https://github.com/kubernetes-sigs/cluster-api-operator/blob/dbd48bb0c3323972a3b15e076f3948291b9f238a/api/v1alpha2/controlplaneprovider_wrapper.go#L50 instead

mjura:
> Thanks for the fix. Possibly this needs to be charged in
> 
> https://github.com/kubernetes-sigs/cluster-api-operator/blob/dbd48bb0c3323972a3b15e076f3948291b9f238a/api/v1alpha2/controlplaneprovider_wrapper.go#L50
> instead

Thank you @Danil-Grigorev for review and fix suggestion. I moved fix to `controlplaneprovider_wrapper.go`.

I have tried to run `make generate`  to see if there will be some changes in API and looks clean.

PR #871, opened 2025-08-08
✨ Implement diagnostics fields in Providers CRD

fariaass:
This PR adds the fields `DiagnosticsAddress` and `InsecureDiagnostics` to `ControllerMetrics` struct. It also configures these fields as flags in the deployment.
As of CAPI v1.9, the flag `--metrics-bind-addr` was removed, so the flag `--diagnostics-address` must be used in place.


furkatgofurov7 (approved):
/approve

PR #876, opened 2025-08-19
:sparkles: Support alter components custom func

salasberryfin:

With the recent changes in CAPI Operator that expose phase reconciliation, it would be great if we can also make applying changes on provider manifest definition easier during one of these phases.

This change introduces a new field in `PhaseReconciler` to embed one/multiple [ComponentsAlterFn](https://github.com/kubernetes-sigs/cluster-api/blob/7782ab73d31d3e1ffe3c46966bd438efe592be72/cmd/clusterctl/client/repository/components.go#L149) that are applied on `Fetch`. With this solution, specific components on a provider's manifest definition can be patched based on the logic passed in this/these function/functions.

Additionally, this adds an extra commit to fix a typo in one of the struct names `PhaseReconcilerOption`. **Looks like this is the reason CI's triggerring api diff issues**, I can drop this change if it delays having the new feature added, and I can open a separate PR to fix the typo.

Fixes #

### DIFF
--- a/controller/alias.go
+++ b/controller/alias.go
@@ -66,3 +66,6 @@ type ProviderMapper = providercontroller.ProviderMapper
 
 // WithProviderMapper is an alias for the internal WithProviderMapper function.
 var WithProviderMapper = providercontroller.WithProviderMapper
+
+// WithCustomAlterComponentsFuncs is an alias for the internal WithCustomAlterComponentsFuncs function.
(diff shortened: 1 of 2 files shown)


mjura (approved):
LGTM

furkatgofurov7 (commented):
/lgtm

PR #893, opened 2025-09-20
:sparkles: Add fetchConfig support for all provider types

### LINKED ISSUE #834: Support for FetchConfig on all provider types in Helm Chart
**User Story**

As an operator, I would like to deploy CAPI Operator Helm Chart with predefined set of providers and be able to specify FetchConfig for all kinds; currently, only infrastructure providers support it.

I believe this logic (in `templates/infra.yaml`) should be extended to the rest of templates
```
{{- if and (kindIs "map" $.Values.fetchConfig) (hasKey $.Values.fetchConfig $infrastructureName) }}
{{- range $key, $value := $.Values.fetchConfig }}
  {{- if eq $key $infrastructureName }}
  fetchConfig:
    {{- range $k, $v := $value }}
      {{ $k }}: {{ $v }}
    {{- end }}
  {{- end }}
{{- end }}
{{- end }}
```

/kind feature

WesleyKlop:

It is currently not possible to specify fetchConfig for all resource types, only for infrastructure and ipam types. This PR adds support to all provider types.

There is one issue though: for example the [k3s/cluster-api-k3s](https://github.com/k3s-io/cluster-api-k3s/) project provides both control-plane and bootstrap providers, under the same name. But since the fetchConfig mapping is not taking into consideration the type of resource, collisions can occur.

I did not want to change the workings of the fetchConfig value too much before discussing but I had several ideas on how to solve this issue:

1. Prefix names with their type. So you could have something like:
   ```yaml
   fetchConfig:
     # Having <provider> as a key and the <name> under it would be fine by me too.
     controlPlane/k3s:
       url: ''
     bootstrap/k3s:
       url: ''
   ```
2. Move fetchConfig values to their provider definition. In which case your config could look like:
   ```yaml
   bootstrap:
     k3s:
       fetchConfig:
         url: ''
   ```

Changing hoe fetchConfig works could also be considered out of scope for this PR, but then we could maybe follow up on it in a different issue.

Fixes #834

### DIFF
--- a/hack/charts/cluster-api-operator/templates/addon.yaml
+++ b/hack/charts/cluster-api-operator/templates/addon.yaml
@@ -45,9 +45,7 @@ spec:
 {{- end }}
 {{- if and (kindIs "map" $.Values.fetchConfig) (hasKey $.Values.fetchConfig $addonName) }}
   fetchConfig:
-    {{- range $key, $value := index $.Values.fetchConfig $addonName }}
-      {{ $key }}: {{ $value }}
-    {{- end }}
+    {{- index $.Values.fetchConfig $addonName | toYaml | nindent 4 }}
 {{- end }}
 {{- if $addon.additionalManifests }}
   additionalManifests:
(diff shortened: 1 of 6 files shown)


furkatgofurov7:
Hey,

#780 is merged now. Can we close this PR?

PR #915, opened 2025-10-21
🐛 Return error when customizing non-existent containers

Fedosin:
Previously, when a user specified container customizations in the provider spec for a container that doesn't exist in the deployment, the operator would silently ignore the configuration. This made it difficult to debug misconfigured provider specs.

This change adds error handling to the `customizeContainer` function to return an error when no matching container is found in the deployment. The error message includes both the container name and deployment name to help users quickly identify and fix the issue.

This makes the error handling consistent with the existing pattern used for manager container validation.

Fixes #


furkatgofurov7 (approved):
/approve

alexander-demicev (approved):
/lgtm

PR #986, opened 2026-02-11
✨ feat: make feature gates additive instead of replacing defaults

andrejbella:

This change makes user specified feature gates merge with existing defaults from the manifest instead of replacing them entirely.

Previously, specifying featureGates would remove all default feature gates from the upstream manifest. Now it preserves existing gates and only overrides the ones explicitly specified.

Fixes #


andrejbella:
sorry, closing as i pushed with wrong account without cla

PR #993, opened 2026-02-27
🐛 Set insecure-diagnostics Arg only when true

SVanhull:

The manager flag `insecure-diagnostics` defaults to `false` but CAPO (https://github.com/oracle/cluster-api-provider-oci) does not use `AddManagerOptions` from cluster-api/util/flags to configure the manager causing a failure to start when `--insecure-diagnostics` is unexpectedly passed. Omitting this Arg when false will allow CAPO (and any other providers that skip this step) to boot while retaining the ability to enable this on other providers that configure the manager with `AddManagerOptions`

Fixes #


furkatgofurov7 (commented):
/lgtm

PR #1010, opened 2026-03-12
✨ feat(chart): core: allow passing additionalArgs to manager

fcuello-fudo:

Allows to pass additionalArgs to the manager, for example, to set --additional-sync-machine-labels or --additional-sync-machine-annotations.


furkatgofurov7 (commented):
Hey, can you please add a test to this change?

PR #1034, opened 2026-04-10
🐛  Allow setting empty tolerations on provider deployments

### LINKED ISSUE #1018: spec.deployment.tolerations: [] does not invalidate provider cache, so upstream tolerations remain

1. Install a provider (for example `BootstrapProvider` or `ControlPlaneProvider`) without setting `spec.deployment.tolerations`.
2. Let the operator reconcile and create the provider Deployment from the upstream manifests.
3. Update the provider and set:

```yaml
spec:
  deployment:
    tolerations: []
```

4. Observe that the provider Deployment / pods still keep the upstream tolerations for control-plane/master nodes.

In my case this shows up as bootstrap/controlplane provider pods still carrying tolerations like:

- `node-role.kubernetes.io/master`
- `node-role.kubernetes.io/control-plane`

even after setting `spec.deployment.tolerations: []`.

I expected `spec.deployment.tolerations: []` to clear the upstream tolerations from the rendered provider Deployment, so the provider pods would no longer tolerate scheduling onto master/control-plane nodes.

I was able to reproduce this with focused regression tests.

Observed behavior from tests:
- `tolerations: []` survives API round-trip and is read back as an explicit empty slice
- deployment customization already supports clearing tolerations when it sees a non-nil empty slice
- however, updating from “tolerations not set” to `tolerations: []` does not change the provider cache hash on current `main`
- as a result, stale cached manifests can be reused and the upstream tolerations remain in place

This appears to be a bug in the cache invalidation / provider hash path rather than in the `DeploymentSpec.Tolerations` API field itself.

/kind bug

tahsinrahman:
Changes `DeploymentSpec.Tolerations` from `[]corev1.Toleration` to `*[]corev1.Toleration` so the three states — unset (nil), explicitly empty (&[]), and non-empty — are distinguishable.

Previously, setting `spec.deployment.tolerations: []` produced the same json.Marshal output as not setting tolerations at all (null vs [] both marshal identically for a bare slice). This meant the provider cache hash did not change, stale cached manifests were reused, and upstream tolerations (e.g. `node-role.kubernetes.io/control-plane`) remained on the provider Deployment.

With the pointer, json.Marshal emits `"tolerations":null` vs `"tolerations":[]`, producing different hashes and triggering a proper re-reconciliation.

Fixes #1018

### DIFF
--- a/internal/controller/component_customizer_test.go
+++ b/internal/controller/component_customizer_test.go
@@ -153,12 +153,12 @@ func TestCustomizeDeployment(t *testing.T) {
 		{
 			name: "only tolerations modified",
 			inputDeploymentSpec: &operatorv1.DeploymentSpec{
-				Tolerations: []corev1.Toleration{
+				Tolerations: ptr.To([]corev1.Toleration{
 					{
 						Key:    "node-role.kubernetes.io/master",
 						Effect: corev1.TaintEffectNoSchedule,
 					},
-				},
+				}),
 			},
 			expectedDeploymentSpec: func(inputDS *appsv1.DeploymentSpec) (*appsv1.DeploymentSpec, bool) {
 				expectedDS := &appsv1.DeploymentSpec{
@@ -345,12 +345,12 @@ func TestCustomizeDeployment(t *testing.T) {
 			inputDeploymentSpec: &operatorv1.DeploymentSpec{
 				Replicas:     ptr.To(3),
 				NodeSelector: map[string]string{"a": "b"},
-				Tolerations: []corev1.Toleration{
+				Tolerations: ptr.To([]corev1.Toleration{
 					{
 						Key:    "node-role.kubernetes.io/master",
 						Effect: corev1.TaintEffectNoSchedule,
 					},
-				},
+				}),
 				Affinity: &corev1.Affinity{
 					NodeAffinity: &corev1.NodeAffinity{
 						RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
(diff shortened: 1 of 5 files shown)


furkatgofurov7:
Hey, this needs a commit message to be fixed (remove mentioning issues, i.e closes https://github.com/kubernetes-sigs/cluster-api-operator/issues/1018)

internal/controller/component_customizer_test.go
```diff
@@ -345,12 +345,12 @@ func TestCustomizeDeployment(t *testing.T) {
 			inputDeploymentSpec: &operatorv1.DeploymentSpec{
 				Replicas:     ptr.To(3),
 				NodeSelector: map[string]string{"a": "b"},
-				Tolerations: []corev1.Toleration{
+				Tolerations: tolerationSlicePtr([]corev1.Toleration{
```
Danil-Grigorev:
```suggestion
				Tolerations: ptr.To([]corev1.Toleration{
```

internal/controller/component_customizer_test.go
```diff
@@ -153,12 +153,12 @@ func TestCustomizeDeployment(t *testing.T) {
 		{
 			name: "only tolerations modified",
 			inputDeploymentSpec: &operatorv1.DeploymentSpec{
-				Tolerations: []corev1.Toleration{
+				Tolerations: tolerationSlicePtr([]corev1.Toleration{
```
Danil-Grigorev:
```suggestion
				Tolerations: ptr.To([]corev1.Toleration{
```

Danil-Grigorev (commented):
Looks good in general, though ptr.To should be used instead.

tahsinrahman:
thanks for your review @Danil-Grigorev @furkatgofurov7 , i've updated

tahsinrahman:
CI is failing due to the api changes. not sure what is the way forward here 

```
sigs.k8s.io/cluster-api-operator/api/v1alpha2
  Incompatible changes:
  - DeploymentSpec.Tolerations: changed from []k8s.io/api/core/v1.Toleration to *[]k8s.io/api/core/v1.Toleration
```

alexander-demicev:
@tahsinrahman hi, can you rebase?

furkatgofurov7:
Hi,

This needs a rebase on top of main HEAD to fix govuln CI failure

furkatgofurov7:
We are planning to cut a new operator release and include this fix in the release.
@tahsinrahman can you please rebase this PR so we can go ahead with merging?

furkatgofurov7 (commented):
/lgtm

PR #1041, opened 2026-04-17
✨ helm: Expose manager verbosity parameter for all provider types

TheGthr:

The provider CRDs support a `manager.verbosity` field but it is not exposed in the Helm chart. The PR adds `verbosity` to the `manager` spec block in all the provider templates, defaulting to `1` when not set (existing behavior).


furkatgofurov7:
Hi,

This needs a rebase on top of main HEAD to fix govuln CI failure

PR #1045, opened 2026-04-29
🌱 Bump go to v1.25.9

Prajyot-Parab:
- Bump go to v1.25.9

Fixes #

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,6 +1,6 @@
 version: "2"
 run:
-  go: "1.25.8"
+  go: "1.25.9"
   allow-parallel-runners: true
 linters:
   default: none
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ ROOT:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
 .DEFAULT_GOAL:=help
 
-GO_VERSION ?= 1.25.8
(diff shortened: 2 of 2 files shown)


damdo:
@Prajyot-Parab

Thanks, changes look reasonable, on top of those, personally,
I'd suggest also adding the `toolchain` directive in the `go.mod`, `test/go.mod`, `hack/chart-update/go.mod` and `hack/tools/go.mod`.

```
go 1.25.8 # leave this as low as possible for compatibility with whomever can only import it/build it with lower versions of go ( so even go 1.25.0 if possible)

toolchain go1.25.9 # bump this as high as possible relative to the 1.25.z version so you instruct the compilers to use the highest available that might have bugfixes.
```

This is the same way we do it in CAPG ([see](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/c0ebb0fce031947d9b2e7e64b015aa7699b3f676/go.mod#L3-L5)).

More details on this: https://medium.com/@alif_biswas/why-your-go-mod-has-two-versions-go-and-toolchain-go-996f8065482e

But better to check with @alexander-demicev @Fedosin @furkatgofurov7 
what they prefer :)

alexander-demicev:
@damdo it makes sense, lets open a follow up PR for toolchain version and merge this one to unblock govulncheck

